PR #21, opened 2022-03-09
Allow use of compile time and runtime parameters on the <T>LAPACK routines inside the directory `lapack`

weslleyspereira:
The advantage of using template types (e.g., `uplo_t`) instead parameters (e.g., `Uplo`) on the interface is that one may call a routine (e.g., `potrf2( uplo_t uplo, matrix_t& A )`) in the following ways:

```C++
uplo = Uplo::Upper;
potrf2( uplo, A ); // Choice of the parameter value at runtime. potrf2 will check the parameter at runtime.

potrf2( Uplo::Lower, A ); // Choice of the parameter value at compile time. potrf2 will check the parameter at runtime.

constexpr upper_triangle_t upper_triangle = { };
potrf2( upper_triangle, A ); // Choice of the parameter value at compile time. potrf2 will check the parameter at compile time.
```

Those 3 calls would call instantiations of the same template. The first two call `potrf2<Uplo,...>` and the last one calls `potrf2<upper_triangle_t,...>`. Note that, to obtain the desired behavior we must define the cast operator inside `upper_triangle_t`:

```C++
struct upper_triangle_t {
    constexpr operator Uplo() const { return Uplo::Upper; }
};
```

Inside `potrf2`, we use:
```C++
if( uplo == Uplo::Upper ) { ... }
```

See also this example using some Norm types: https://godbolt.org/z/3YM9z4356

This PR adds:
- the possibility of using runtime and compile time parameters on all <T>LAPACK routines inside the directory `lapack`.

Missing:
- Routine `lascl`.
- Add error checks to the routines.

### DIFF
--- a/include/blas/arrayTraits.hpp
+++ b/include/blas/arrayTraits.hpp
@@ -11,15 +11,6 @@
 
 namespace blas {
 
-    /// Layouts
-    enum class Layout { ColMajor = 'C', RowMajor = 'R' };
-    struct ColMajor_t {
-        constexpr operator Layout() const { return Layout::ColMajor; }
-    };
-    struct RowMajor_t {
-        constexpr operator Layout() const { return Layout::RowMajor; }
-    };
-
     /// Data type
     template< class T > struct type_trait {};
     template< class T >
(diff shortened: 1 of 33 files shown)


weslleyspereira:
Routine `lascl` is treated separately on #22.

weslleyspereira:
Error checks were added on 992570d.
Also, `lansy` and `lanhe` were improved in 1277f77 following the strategy applied on `lange`.

PR #22, opened 2022-03-09
Introduces access policies and the legacy banded matrix

weslleyspereira:
This commit adds one new feature, called **Access policies**.

Every routine in \<T\>LAPACK assumes access the whole or parts of their input matrices. We call it **access type**. For example, `potrf` accesses either the upper triangle or lower triangle. In `syrk`, one of the matrices has all its positions read, while the other matrix will be read and written only on its upper or lower triangle. Some routines will have input flags that will determine the access type.

**Access policies** define the pairs `(i,j)` where `A(i,j)` returns a valid value (`read_policy()` or reference (`write_policy()`). This is the way guarantees a certain array is allowed for the routine using it. For example, in a banded matrix, the algorithms can only write on the entries inside the band. In a upper triangular matrix, only the upper triangle entries can be accessed. This is critical when the layout prohibits reading and/or writing in certain positions. For instance, the LAPACK packed format only allows for reading and accessing the upper or lower triangle part of the matrix.

The objective of this PR is to introduce a way to verify if the access type is compatible with the data access policy. 
    
Policies and types will have the same range of values. The current access policies are values of

```c++
/**
 * @brief Matrix access policies
 * 
 * They are used:
 * 
 * (1) on the data structure for a matrix, to state the what pairs (i,j)
 * return a valid position A(i,j), where A is a m-by-n matrix.
 * 
 * (2) on the algorithm, to determine the kind of access
 * required by the algorithm.
 */
enum class MatrixAccessPolicy {
    Dense = 'G',            ///< 0 <= i <= m,   0 <= j <= n.
    UpperHessenberg = 'H',  ///< 0 <= i <= j+1, 0 <= j <= n.
    LowerHessenberg = 'h',  ///< 0 <= i <= m,   0 <= j <= i+1.
    UpperTriangle = 'U',    ///< 0 <= i <= j,   0 <= j <= n.
    LowerTriangle = 'L',    ///< 0 <= i <= m,   0 <= j <= i.
    StrictUpper = 'S',      ///< 0 <= i <= j-1, 0 <= j <= n.
    StrictLower = 's',      ///< 0 <= i <= m,   0 <= j <= i-1.
};
```

and objects of the class

```c++
/**
 * @brief Band access
 * 
 * Pairs (i,j) such that max(0,j-ku) <= i <= min(m,j+kl) in a m-by-n matrix,
 * where kl is the lower_bandwidth and ku is the upper_bandwidth.
 * 
 *      x x x 0 0
 *      x x x x 0
 *      0 x x x x
 *      0 0 x x x
 */
struct band_t : dense_t {
    std::size_t lower_bandwidth; ///< Number of subdiagonals.
    std::size_t upper_bandwidth; ///< Number of superdiagonals.

    constexpr band_t(std::size_t kl, std::size_t ku)
    : lower_bandwidth(kl), upper_bandwidth(ku)
    {}
};
```

One may also use the following constant expressions instead of values of `MatrixAccessPolicy`:

```c++
// constant expressions
constexpr dense_t dense = { };
constexpr upperHessenberg_t upperHessenberg = { };
constexpr lowerHessenberg_t lowerHessenberg = { };
constexpr upperTriangle_t upperTriangle = { };
constexpr lowerTriangle_t lowerTriangle = { };
constexpr strictUpper_t strictUpper = { };
constexpr strictLower_t strictLower = { };
```

The use of constant expressions enable compile time checking.

### How does the verification between access policy and access policy work?

We propose to use two checkers: `access_granted` and `access_denied`, to do the comparison. The latter is the negative of the former. The former is as follows:

```c++
/**
 * @brief Check if a given access type is compatible with the access policy.
 * 
 * Examples of outputs:
 * 
 *      access_granted( MatrixAccessPolicy::UpperTriangle,
 *                      MatrixAccessPolicy::Dense )             returns true.
 *      access_granted( MatrixAccessPolicy::Dense,
 *                      MatrixAccessPolicy::LowerHessenberg )   returns false.
 *      access_granted( Uplo::Upper,
 *                      MatrixAccessPolicy::UpperHessenberg )   returns true.
 *      access_granted( Uplo::Upper,
 *                      Uplo::Lower )                           returns false.
 *      access_granted( upperTriangle,
 *                      Uplo::Lower )                           returns false.
 *      access_granted( strictUpper,
 *                      Uplo::Upper )                           returns true.
 * 
 * @tparam access_t         Access type.
 *      Either MatrixAccessPolicy, Uplo, or any type that implements
 *          operator MatrixAccessPolicy().
 * @tparam accessPolicy_t   Access policy.
 *      Either MatrixAccessPolicy, Uplo, or any type that implements
 *          operator MatrixAccessPolicy().
 * 
 * @param a Access type.
 * @param p Access policy.
 * 
 * @ingroup utils
 */
template< class access_t, class accessPolicy_t >
inline constexpr
bool access_granted( access_t a, accessPolicy_t p ) { /* ... */ }

/**
 * @return ! access_granted( a, p ).
 * 
 * @see bool access_granted( access_t a, accessPolicy_t p )
 * 
 * @ingroup utils
 */
template< class access_t, class accessPolicy_t >
inline constexpr
bool access_denied( access_t a, accessPolicy_t p ) {
    return ! access_granted( a, p );
}
```

The rules for granting access are based on the the diagram of inheritance:

structlapack_1_1dense__t__inherit__graph

In a few words, an access type that derives from an access policy will result in "access granted".

### Access policy is different from layout in at least two ways:

1.  A upper triangular matrix can be stored using either a dense or a
    packed layout, but the access policy will always be: `A(i,j)`
    for `i <= j`.

2.  The access policy has to do with the algorithms, so we suppose there
    is a small number of ways an algorithm will access a given data. For
    example, in LAPACK, we have: dense, upper/lower Hessenberg,
    upper/lower triangle, upper/lower strict triangle, and banded. The
    layout has to do with specialized code. For example, for the
    column-major layout, one may use BLAS algorithms from optimized BLAS
    packages like MKL and OpenBLAS.
       
------------
    
This PR also:

- Creates `legacyBandedMatrix` and and the helper `function banded_matrix()`.
- Contains an example for the use of access policies: `example/access_types`.
- New (optional) trait: `layout_trait`. This simplifies the layout checks from `allow_optblas`. 
- Use of the concept Argument-Dependent Lookup (ADL) present in C++ to avoid ambiguous calls of overloaded functions.
- Provides major improvement in the design of `lascl`.
- Has some bug fixes.


weslleyspereira:
Description above was updated to better represent the changes.

PR #27, opened 2022-03-27
Implement the idea from @thijssteel on #26

weslleyspereira:
Use `slice` to represent the block operations on both matrices and vectors like proposed on #26. It considerably reduces the wordiness of the code. I personally like the way the code looks now. Note that I maintained the block operations: `row`, `rows`, `col` and `cols`. Any other suggestion on this thread, @thijssteel ?


thijssteel:
no, looks good

thijssteel:
This sparked the idea for PR #32, but while we discuss that further, any chance we could get this merged? It is used in a lot of other functions, so getting it into the main branch as soon as possible would avoid merge conflicts.

PR #28, opened 2022-03-27
Fix workspace in larf and larfb.

weslleyspereira:
It is unusual to throw any error if the workspace is larger than the necessary. This PR fixes `larf` and `larfb`. Now all the routines shall accept workspace larger than the necessary. Thanks to @thijssteel for finding this issue!


weslleyspereira:
The issue on `larf` is fixed on #39. I will wait for #39 before pushing these changes.

PR #32, opened 2022-03-30
Routines may accept vectors or matrices

weslleyspereira:
This PR proposes a simple way to allow the seamless use of vectors as 1-column matrices. I test this functionality with the function `lange`, and use a new example to verify the proposal works.

If this proposal is satisfactory, we could reduce considerably the number of routines in \<T\>LAPACK. The matrix version of each routine would be the preferable one, e.g., `gemm`, `hemm`, `trmm` and `trsm`. Routines like `gemv`, `hemv`, `trmv` and `trsv` could be deprecated. Moreover, routines like `lange`, `nrm2` and `asum` could be easily integrated into a single one. The goal is to increase maintainability and reduce the number of functions, which also ease the use of the library. The modifications should not change the overall performance of the code.

### DIFF
--- a/examples/interpretAsMatrix/CMakeLists.txt
+++ b/examples/interpretAsMatrix/CMakeLists.txt
@@ -1,0 +1,20 @@
+# Copyright (c) 2021-2022, University of Colorado Denver. All rights reserved.
+#
+# This file is part of <T>LAPACK.
+# <T>LAPACK is free software: you can redistribute it and/or modify it under
+# the terms of the BSD 3-Clause license. See the accompanying LICENSE file.
+
+cmake_minimum_required(VERSION 3.1)
+
+project( interpretAsMatrix CXX )
+
+# Load <T>LAPACK
+if( NOT TARGET tlapack )
+  find_package( tlapack REQUIRED )
+endif()
+
+find_package( mdspan REQUIRED )
+
+# add the example interpretAsMatrix
+add_executable( example_interpretAsMatrix example_interpretAsMatrix.cpp )
+target_link_libraries( example_interpretAsMatrix PRIVATE tlapack std::mdspan )
--- a/examples/interpretAsMatrix/example_interpretAsMatrix.cpp
+++ b/examples/interpretAsMatrix/example_interpretAsMatrix.cpp
@@ -1,0 +1,49 @@
+/// @file example_interpretAsMatrix.cpp
+/// @author Weslley S Pereira, University of Colorado Denver, USA
+//
+// Copyright (c) 2021-2022, University of Colorado Denver. All rights reserved.
+//
+// This file is part of <T>LAPACK.
+// <T>LAPACK is free software: you can redistribute it and/or modify it under
+// the terms of the BSD 3-Clause license. See the accompanying LICENSE file.
+
+// #include <tlapack.hpp>
+
+#include <vector>
+#include <iostream>
+
+#include <plugins/tlapack_mdspan.hpp>
+#include <tlapack.hpp>
+
+//------------------------------------------------------------------------------
+int main( int argc, char** argv )
+{
+    using T = float;
+    using idx_t = std::size_t;
+
+    using namespace lapack;
+    using std::experimental::mdspan;
+    using std::experimental::extents;
+    
+    // constants
+    const idx_t n = 10;
+    
+    // raw data arrays
+    T v_[ n ] = {1,2,3,4,5,6,7,8,9,10};
+    
+    // print array
+    for (idx_t i = 0; i < n; i++)
+        std::cout << v_[i] << ", ";
+    std::cout << std::endl;
+
+    // mdspan arrays
+    mdspan< T, extents<10,1> > A{ v_ };
+    mdspan< T, extents<1,10> > B{ v_ };
+    mdspan< T, extents<10> > v{ v_ };
+
+    std::cout << lange( one_norm, A ) << std::endl;
+    std::cout << lange( one_norm, B ) << std::endl;
+    std::cout << lange( one_norm, v ) << std::endl;
+
+    return 0;
+}
--- a/include/lapack/lange.hpp
+++ b/include/lapack/lange.hpp
@@ -36,15 +36,17 @@ namespace lapack {
  * 
  * @ingroup auxiliary
 **/
-template< typename norm_t, typename matrix_t >
-real_type< type_t< matrix_t > >
-lange( norm_t normType, const matrix_t& A )
+template< typename norm_t, typename array_t >
+auto
+lange( norm_t normType, const array_t& a )
 {
-    using T      = type_t< matrix_t >;
+    using T      = type_t< array_t >;
     using real_t = real_type< T >;
-    using idx_t  = size_type< matrix_t >;
+    using idx_t  = size_type< array_t >;
     using blas::isnan;
     using blas::sqrt;
+
+    const auto& A = interpretAsMatrix( a );
 
     // constants
     const real_t rzero(0.0);
--- a/include/plugins/tlapack_abstractArray.hpp
+++ b/include/plugins/tlapack_abstractArray.hpp
@@ -240,6 +240,29 @@ namespace blas {
     inline constexpr auto
     diag( const matrix_t& A, idx_t diagIdx = 0 );
 
+    /**
+     * @brief Interpret a given vector as a 1-column matrix.
+     * 
+     * @tparam vector_t Vector type.
+     * 
+     * @param v Vector of size n.
+     * 
+     * @return Matrix of size n-by-1.
+     */
+    template< class vector_t >
+    inline constexpr
+    auto
+    interpretAsMatrix( const vector_t& v );
+
+    /// @returns The input because it is already a matrix.
+    template< class matrix_t >
+    inline constexpr
+    const auto&
+    interpretAsMatrix( const matrix_t& A )
+    {
+        return A;
+    }
+
     // -------------------------------------------------------------------------
     // Block operations with vectors in <T>LAPACK
 
--- a/include/plugins/tlapack_eigen.hpp
+++ b/include/plugins/tlapack_eigen.hpp
@@ -229,10 +229,34 @@ namespace blas{
     {
         return A.diagonal( diagIdx );
     }
-    template<typename XprType, int BlockRows, int BlockCols, bool InnerPanel, typename SliceSpec, class int_t>
+    template<typename XprType, int BlockRows, int BlockCols, bool InnerPanel, class int_t>
     inline constexpr auto diag( Eigen::Block<XprType, BlockRows, BlockCols, InnerPanel>&& A, int_t diagIdx = 0 ) noexcept
     {
         return A.diagonal( diagIdx );
+    }
+
+    // interpretAsMatrix
+    template<class T>
+    inline constexpr
+    const Eigen::MatrixBase<T>&
+    interpretAsMatrix( const Eigen::MatrixBase<T>& A ) noexcept
+    {
+        return A;
+    }
+    
+    template<class T>
+    inline constexpr
+    Eigen::MatrixBase<T>&
+    interpretAsMatrix( Eigen::MatrixBase<T>& A ) noexcept {
+        return A;
+    }
+    
+    template<typename XprType, int BlockRows, int BlockCols, bool InnerPanel>
+    inline constexpr
+    auto&&
+    interpretAsMatrix( Eigen::Block<XprType, BlockRows, BlockCols, InnerPanel>&& A ) noexcept
+    {
+        return std::forward(A);
     }
 
 } // namespace blas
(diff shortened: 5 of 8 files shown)


langou:
I agree with @thijssteel that I am not convinced about the argument "reduce considerably the number of routines". 

My main issue is that I want to make sure that if we solve Ax=b, then it works for x being a 1D array and x being a 2D array. If we support just one, we should support the 2D array. But it would be nice that getrs.hpp can support both. I think this is the main point of this PR.

weslleyspereira:
I agree with you. When I said "reduce considerably the number of routines" I was thinking on the current code on the library, which has a lot from BLAS. The idea is to design routines for matrices, and then have them working on vectors too.

weslleyspereira:
I think we can accept vectors as matrices at least in some routines like: `trsm`, `lange` and `gesv`. In these cases we would:

- avoid writing `trsv`;

- Homogenize the ways of calculating norms for vectors and matrices;

- Have `gesv` working for matrices and vectors. Notice that we would need anyway to write 2 overloads for `gesv`.

I want to share an alternative to [6a158d3](https://github.com/tlapack/tlapack/pull/32/commits/6a158d3af43df4b8ec3e525ee9325ca063965fcb) to accomplish the same behavior. Two advantages of this new approach:

- No need to define any function like `interpretAsMatrix`. Instead, we need a trait `is_matrix< array_t >` that can be either enforced via specialization or automatically predicted. We also need the trait `is_vector< array_t >` for vectors to make the code recognize true vectors.

- This new version is supposed to be better in terms of performance. This is because, at compile time: the number of columns is set to 1; and the access discard the column index.

This is what we have of new code:

1. New traits `is_matrix< array_t >` and `is_vector< array_t >`:

```C++
// Possible implementations of the automatic prediction:

template< class array_t >
constexpr bool is_matrix = !is_same_v< layout_type<array_t>, void >;

template< class array_t >
constexpr bool is_vector = false;

// In the plugin `tlapack_stdvector.hpp` :
template< class T, class Allocator >
constexpr bool is_vector< std::vector<T,Allocator> > = true;
```

2. New structure `oneColMatrix`:

```c++
template< class vector_t >
struct oneColMatrix : public vector_t {
    
    using idx_t = size_type< vector_t >;
    using T     = type_t< vector_t >;

    inline constexpr T&
    operator()( idx_t i, idx_t j ){ return (*this)[i]; }
    
    inline constexpr T
    operator()( idx_t i, idx_t j ) const { return (*this)[i]; }
};

// is_matrix< array_t > can be specialized if one data structure does not define layout trait:
template< class vector_t >
constexpr bool is_matrix< oneColMatrix<vector_t> > = true;

template< class vector_t >
struct type_trait< oneColMatrix<vector_t> > {
    using type = type_t< vector_t >;
};

template< class vector_t >
struct sizet_trait< oneColMatrix<vector_t> > {
    using type = type_t< vector_t >;
};

template< class vector_t >
inline constexpr auto
nrows( const oneColMatrix<vector_t>& A ){ return size(A); }

template< class vector_t >
inline constexpr auto
ncols( const oneColMatrix<vector_t>& A ){ return 1; }

template< class vector_t >
inline constexpr auto
read_policy( const oneColMatrix<vector_t>& A ) { return dense; }

template< class vector_t >
inline constexpr auto
write_policy( const oneColMatrix<vector_t>& A ) { return dense; }
```

3. Routine `foo` has 1 interface with 2 overloads: one contains the algorithm, the other is a wrapper.

```c++
/**
 * Implementation for matrices
 */
template< typename norm_t, typename matrix_t,
    // We need an extra check:
    enable_if_t< is_matrix< matrix_t >, int > = 0
>
auto
foo( norm_t normType, const matrix_t& A )
{    
    using T      = type_t< matrix_t >;
    using idx_t  = size_type< matrix_t >;

    // constants
    const idx_t m = nrows(A);
    const idx_t n = ncols(A);

    return (m > 2) ? A(2,0) : -1;
}

/**
 * Wrapper for vectors
 */
template< typename norm_t, typename vector_t,
    // We need an extra check:
    enable_if_t< !is_matrix< vector_t >, int > = 0
>
inline auto
foo( norm_t normType, const vector_t& v )
{    
    return foo( normType, static_cast< const oneColMatrix<vector_t>& >(v) );
}
```
How one will use it:

```c++
int main() {
    std::vector<float> v = {1,2,3,4,5,6};
    return (int) foo( 'F', v );
}
```

Complete [code on godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAMzSrgAyeAyYAHI%2BAEaYxBKBpAAOqAqETgwe3r4B0ilpjgKh4VEssfFciXaYDhlCBEzEBFk%2BflWY9oUM9Y0ExZExcQnSCg1NLTlVY31hA2VDlZIAlLaoXsTI7Bzm/mHI3lgA1Cb%2BbgQAnkmYAPoExEyECifYJhoAgjt7B5jHpwBuNSIxGerw%2B7wA9OCALQw2Fw%2BEIxFI5Eo1Gw0GQw5vWjETBMdDnQ5hX5uAAqz2Cb2UbzcAGkQKDQV40kZDqN0CAQIwmNF6Nc8FRbicrO8CJgWEkDGKTm59kwFApDqTSHKFYc5CD3mgGKNMKoksRDtFUJ4iQprgpWDc/r8ACJsggckB4c2Wtgy5Ua/zYTl/MReTDCxmi8WSpjS06HVWKxr3c63X7YB3ELwOQ4XK63e6EY4AdgsedtQZDEqlgcj0cOsaY8YIieTqbraQAXpgCFmHnWTPnC8W3mLS%2BHy24owY1dXa/XRim0wZzmt23dO3mRW9DuvDsywsB05cfid7X9UHh0MLDpDwTLDrbMDRwoqCAgfnOF7urgA6UEb78/38/i9XgAav6PxHiehxsIYirmGYwQ1q%2BDCoHWYT8MQbCnmYZifu83ZFv4q4Ypi2J4PKhyoYcAACyBvjcS6EDKzychmmDYf2oZllelYTgmmpvFurLMTx/j2sxzBsDRHb0ZG3Fdt6TF7n2hHglitAkYq5FUWyeCtou2YEAxckgMxrEDmGEYjlxxBxkJLzvPxO4trRe52jRYk/I5umdleMmJvJVyKRCRGqaRGnUS%2BXieVJbiMUZe4mexQ6cWOiqkomTIsju4WLs5B6uVahxZZJ%2BmRqlMXMQFbzTo2hxYDqtErueABU74tY1ym4X22q6vqhq1UoCZ9fuwkNR1%2BHBm8kJolN00zTNhGHFw76HBEmAAO7pnpIBmtcLDhsQeCqN5Vk1jZ42mRxFbJVWx2TrxXVij1RomrQ227XcB0uWAYAuhaVrXH8V6FeVpwyc8pCHGB6BpWNOElmZw6jvKMY3adWoCN1BpPaaP0Ag4JAuVQYhKBVk2zWT5Nk/NHoUlSNK0tdcabhlDpOrjQLU96ZH44%2BPx6qwkrsGdCXmYjarKqLirYrQqCiECaVozqD2Y8a2PmmzJBXuyvqAhrpzKlLMvhrrSbPC5dwBhV52JZdSNKuDlYG7L%2BO8VVaaCXRxUjlrIDq8CeukI7Rt%2BybnPdqu372TRLnkjDHy5nhBFwxdFlXeLDu0NLTuGi75tph5RWa462t437ZIBxnhvs6Hof5l%2BG6R8xZt7m5LPF1XZeB1XPogI5twKbHo2J2xg4i5WadXZ3zveqCYSqeEUbo0rhpMBFqCgo5EAL4rrc%2Bzrpf6xXWcgmYABshyqIcSwrnX664gQ6wMOf74b0sinx%2BNpMU1/3%2BIvNWHLWtBsDh1jsEOAITAWQACye0DpCxHgjSsvtUaVVznWcBUCYEXy2kkLwvI8DUSQV2Wu7xvw3yZtuIk6BVBCXtL3YGI5CHQ3DvXZmqVSHDXdleRhIIB4kI3LPWYW8MaGnJKfMhqArj3CBBAJYm8TzUOQuDeRCZtCXzDocO%2BD9DgQEao%2BF0r8ACsFg8AmAMQnXsfD1xkIEfPe6epMbkksWAyRQcZFyKoQmPASiPF1lUVfe6DVNHEEfjovRChDHGNMeYjqOF368ImspH6b19qHWkijWSSZRCP1iGyK4yASKqVbFDAUYD57oHDEwIB99cQ1VQJgRUiE6xYDvM%2BeCEUNqdgZEna2KdbbcOngrYRWMXpJMwVedBnhoHvVSW4Qhpt5kcJTOWIeVtR5XX6bZFBM46zuz0uM8IGDpkyjmdXE2xC1wsIoY3XKnDIwbLfgnOB8Mkp9L3sg12TZtJtgLpGCZtApkpOOW84qIcznMPXA3HKHC9xCjucCphjI4lD0xKUcoYCqAaNQKtJ43S1mvJLsgmxPw7GPRXkQUEDBiBYoUJvAJfyAWwP%2BMC4%2BZ81yvx7EEx%2BG83iv3whYsECTlqDENKgDFaBsg6lBKshB6z4W8SJUIpeVZV4UvFTSxVpSIGTLGUyglILzCsrURytsWiuBnhiQKzEAAlPEUMUiqWQOcKVwsZX4qBIShgc9iWL3scvFV7xcT4muPa/B5xaWL01YcwFur3X6tPliS%2BgSTXBJqowYmfKLXzQAOr7TFIcENjrnXwJeWqe57wFUksxmSte7xVq5puAWsNGr6U6tmcy6e8a2VJuqY/Qa5r36wwST/YdI75r%2BCWm4QNeaeaHDrUwJIkiuaGl9goLplrGqNS/I1Q4ABJUs4pGANC6EuiCMDNg4rXG1ItzzIyiXyohNCtxwZ3vEskg6T6yGYizT8cImAoaGEOHqJcUYnzIAANZru/NyXkNwBSwpHKMo5kY30KMTEowQ9ZcoaEZAMt41bQT8FQJvB9LAEwkdJHue2EaUNCk7Wo2JBYNxkMjmw9hIkYWe1PdM06scI7M2UXWFydD%2B4jhoxkiq35MT3UMAQC934AkCYgi5Sl1KIA8r7PJiNinH65QYGqtTvKh7fk5dolg9YzBXxOAAMSxBAMwpANBXy2lCM1A8B3rs3SQ7dOb52LvIiurdl5cUIxfT%2BkgpGCDPubvlQhpBP3KW/YcX9/7H5AfuCBmoEGyHQb5HBzjOM95cLld6dDGSXLYZwrhhV%2BH3iEeI%2BFsj4WKNXCo9vOZ8abSvwY6QpxJm6tJaa5Rh04Z8HXFEKMTiEaW1IbbXqlliYIB/CvoZxFRZB2fxHZtymgVDgAAksWatnXQF6zIfiEDXTPDDu0wgyOvk472vsZRUGluGU2Npcphy4KQez/hSDSAMaQE%2Bg8yEmYgGEAgV9%2BtgA4FZ6H4NOsPI4CsWgnADG8D8BwLQpBUCcGipYawbI1gbH3GYfwPBSAEE0EjlYEGDFYSkBoLgBiT4aAABwnwMbmLguZ/An30JwSQ6OqfY84LwVdGgKdU5WHAWAMBEAoFQBKOgcRyCUDQEr%2Bg8RgAKGYEkBQCAkJ8DoGKYgq6IDRGF9EMIjRzicHJ%2BrtgggADynrbeY94FgXaRhxDu9IPgXEtQASrt93qGoEUthY/Bx0YXqloj3GIOcDwWA7e8HeiwFPKxntMG14BPAa0ndXAx%2BT/gggRBiHYFIGQghFAqHUL73QX2DBGBQNYaw%2Bg8DRFXZAFYEiujB6hOyA8ph8eWCkIcKETusfVFqM4CArgJh%2BC%2ByEWYqKFjJFSOkAQC%2B9D5E3wwfoq%2BKi2A6ICOo0xt9fen10HoTQD/Cr0KMXoF/bDTDv/MCoKwFBE82HoO4mAtg8DI6o5C6%2B444cCqDs5QgnySCHDADIDUQQDmwMBgZXwQB45WCWDgy4CEDOyk5faHAeAa5xDHB4FLCp5S404gCSAnzvi5is4ACcuYHOrOuYjOkgZgkgjeAuvA6eXO74kgLBGg/g9BzOgQ/guY9Bv2GOWOYBYuIAEulO7u0ucuEASAC4OCkWFAc%2B%2BAQIJ4egJewgog4gleBhNeagwuDepAc6SQGe/OHAaOpA0hvAYBTuEUGh6K58kB0BsB8B2iSBKB2ihBSQyu2cpB5BShKwT4%2BIQwMidhgupA6eZgrO74BiXAZg9BLOkgvOzOZgGguYAOThIuHAchChFBpAEGkguY74rOBi9BrOWR9BiwSRSRdh/gIBMhoukuERdhZg7RzhnRihWgSwKwAIpuGQVBQAA)

PR #34, opened 2022-03-30
Unblocked nonsymmetric eigenvalue routines

thijssteel:
This PR adds unblocked routines to solve eigenvalue problems:
- hessenberg reduction
- generation of orthogonal hessenberg factor
- qr algorithm and some help routines
- an example file to test the implementation slightly

I didn't implement the eigenvector or driver routines yet, that will be for a next PR (i don't want to wait too long an have too many merge conflicts).

Some design choices:
- Instead of storing complex eigenvalues of real matrices in two real arrays wr and wi, i always use complex arrays. This way, the complex interface is the same as the real interface.
- I always use double shifts, this is similar to the multishift implementation in LAPACK, but `zlahqr` does use single shifts. For as far as I know, there is not really a downside.
- The exit check is slightly different for real and imaginary numbers, i don't know a good way to dynamically check the type, so i used overloading on lahqr_schur22, it's not the prettiest solution, so this may need to be changed.

- inside lahqr, i need a fixed size vector workspace (size 3). I just declared this as `std::vector v(3)` i don't know if this will be dynamically allocated, i don't know any other good way other than requesting an input workspace (but that seems dumb).


weslleyspereira:
The error that the Github is pointing here is really awkward.

```cmake
 /Users/runner/work/tlapack/tlapack/examples/gehd2/example_gehd2.cpp:154:65: error: 'upper_triangle' is not a member of 'lapack'; did you mean 'upperTriangle'?
  154 |         norm_orth_1 = lapack::lansy( lapack::frob_norm, lapack::upper_triangle, work );
      |                                                                 ^~~~~~~~~~~~~~
      |                                                                 upperTriangle
```

The name `upperTriangle` was introduced in a recent commit to the master.

Your branch builds as expected in my machine. I am reviewing it.

include/lapack/orghr.hpp
```diff
@@ -0,0 +1,98 @@
...
+    lapack_error_if( (idx_t) size(tau)  < std::min<idx_t>( m, n ), -2 );
+    lapack_error_if( (idx_t) size(work) < n-1, -3 );
+
+
+    // Shift the vectors which define the elementary reflectors one
+    // column to the right, and set the first ilo and the last n-ihi
+    // rows and columns to those of the unit matrix
+
+    // This is currently optimised for column matrices, it may be interesting
+    // to also write these loops for row matrices
```
weslleyspereira, line 61: I want to comment on that. No action required. But if we want to specialize this routine for row-major matrices, we just need to replace

```c++
template<
    class matrix_t, class vector_t, class work_t >
int orghr( ... ) { ... }
```

by 

```c++
template<
    class matrix_t, class vector_t, class work_t,
    enable_if_t<
        ! is_same_v< layout_type<matrix_t>, RowMajor_t >
    , int > = 0
>
int orghr( ... ) { ... }
```

and add a new routine for row major:

```c++
template<
    class matrix_t, class vector_t, class work_t,
    enable_if_t<
        is_same_v< layout_type<matrix_t>, RowMajor_t >
    , int > = 0
>
int orghr( ... ) { ... }
```

include/lapack/lahqr.hpp
```diff
@@ -0,0 +1,419 @@
...
+     *            successfully computed.
+     *
+     * @param[in] want_t bool.
+     *      If true, the full Schur factor T will be computed.
+     * @param[in] want_z bool.
+     *      If true, the Schur vectors Z will be computed.
+     * @param[in] ilo    integer.
+     *      Either ilo=0 or A(ilo,ilo-1) = 0.
+     * @param[in] ihi    integer.
+     *      The matrix A is assumed to be already quasi-triangular in rows and
```
weslleyspereira, line 47: What is the definition of a quasi-triangular matrix? Maybe we should point out to a reference.

include/lapack/lahqr.hpp
```diff
@@ -0,0 +1,419 @@
...
+        using blas::abs;
+        using blas::abs1;
+        using blas::conj;
+        using blas::imag;
+        using blas::uroundoff;
+        using lapack::lahqr_eig22;
+        using lapack::lahqr_shiftcolumn;
+
+        // constants
+        const real_t rone(1);
```
weslleyspereira, line 84: `rone` is not used

include/lapack/lahqr.hpp
```diff
@@ -0,0 +1,419 @@
...
+                    s2 = s1;
+                else
+                    s1 = s2;
+            }
+
+            // We have already checked whether the subblock has split.
+            // If it has split, we can introduce any shift at the top of the new subblock.
+            // Now that we know the specific shift, we can also check whether we can introduce that shift
+            // somewhere else in the subblock.
+            // TODO: Check if this vector leads to allocations
```
weslleyspereira, line 292: I would be surprised if it does. But thanks for commenting on that.

weslleyspereira: You meant "dynamic allocations", right? I understood dynamic allocations.

include/lapack/lahqr_eig22.hpp
```diff
@@ -0,0 +1,78 @@
...
+     *      Element (1,1) of A.
+     * @param[out] s1
+     * @param[out] s2
+     *      s1 and s2 are the eigenvalues of A
+     *
+     * @ingroup geev
+     */
+    template <
+        typename T,
+        typename real_t = real_type<T>>
```
weslleyspereira, line 39: Interesting way of defining `real_t`! I will try to adopt it when dealing with `T` and `real_type<T>` on the function signature.

include/lapack/lahqr_schur22.hpp
```diff
@@ -0,0 +1,222 @@
...
+        return 0;
+    }
+
+    template <
+        typename T,
+        enable_if_t<is_complex<T>::value, bool> = true>
+    int lahqr_schur22(T &a, T &b, T &c, T &d, T &s1, T &s2, T &cs, T &sn)
+    {
+        return -1;
+    }
```
weslleyspereira, line 218: I think we can remove this. If we keep this routine, `lahqr_schur22` will compile with complex types, and return `-1` at runtime. If we remove it, the error will be generated at compile time, which I think it is preferable.

weslleyspereira:
I must mention that the code looks great! Thanks for the contribution @thijssteel !

langou:
Opinion. The name of the PR ``Unblocked sep routines`` is confusing for me. For me, SEP stands for Symmetric Eigenvalue Problem. Here we are doing NEP: Nonsymmetric Eigenvalue Problem. I think SEP stands for ``Solve Eigenvalue Problem`` but  I want ti say that this is confusing to me. LAPACK uses SEP: https://www.netlib.org/lapack/lug/node30.html and NEP: https://www.netlib.org/lapack/lug/node31.html
Maybe SNEP: Solve Nonsymmetric Eigenvalue Problem. Or NEP.

thijssteel:
@langou To me, SEP usually means standard eigenvalue problem to differentiate it from generalized eigenvalue problems. I'll rename the PR.

include/lapack/lahqr.hpp
thijssteel: A quasi-triangular matrix is upper triangular, but can have 2x2 blocks on the diagonal for imaginary eigenvalues in real matrices.
I guess it would be useful to have a specific description of what we mean by standard form.

include/lapack/lahqr_schur22.hpp
thijssteel: lahqr actually needs this. It catches the error to know that it shouldn't treat the 2x2 block as deflated and continue to apply shifts. I couldn't find a cleaner way other than writing lahqr twice.

weslleyspereira: I think I saw the problem. My suggestion is:

```c++
diff --git a/include/lapack/lahqr.hpp b/include/lapack/lahqr.hpp
index e1d888f..4b4a9c2 100644
--- a/include/lapack/lahqr.hpp
+++ b/include/lapack/lahqr.hpp
@@ -214,7 +214,7 @@ namespace lapack
                     istart = ilo;
                     continue;
                 }
-                if (istart + 2 == istop)
+                if ( !is_complex<TA>::value && (istart + 2 == istop) )
                 {
                     // 2x2 block, normalize the block
                     TA cs;
diff --git a/include/lapack/lahqr_schur22.hpp b/include/lapack/lahqr_schur22.hpp
index b24b7bd..a65e1f3 100644
--- a/include/lapack/lahqr_schur22.hpp
+++ b/include/lapack/lahqr_schur22.hpp
@@ -50,9 +50,7 @@ namespace lapack
      *
      * @ingroup geev
      */
-    template <
-        typename T,
-        enable_if_t<!is_complex<T>::value, bool> = true>
+    template < typename T >
     int lahqr_schur22(T &a, T &b, T &c, T &d, std::complex<T> &s1, std::complex<T> &s2, T &cs, T &sn)
     {
 
@@ -74,6 +72,9 @@ namespace lapack
         const T safmn2 = pow(two, (int)(log(safmin / eps) / log(two)) / two);
         const T safmx2 = one / safmn2;
 
+        // Quick return
+        if( is_complex<T>::value ) return -1;
+
         if (c == zero)
         {
             // c is zero, the matrix is already in Schur form.
@@ -209,14 +210,6 @@ namespace lapack
         return 0;
     }
 
-    template <
-        typename T,
-        enable_if_t<is_complex<T>::value, bool> = true>
-    int lahqr_schur22(T &a, T &b, T &c, T &d, T &s1, T &s2, T &cs, T &sn)
-    {
-        return -1;
-    }
-
 } // lapack
 
 #endif // __LAHQR_SCHUR22_HH__
```

You can also use `lapack_error_if( is_complex<T>::value, -1 )` instead of `if( is_complex<T>::value ) return -1`, and this will change slightly the documentation of the method.

thijssteel: I agree that checking for complex values in lahqr is better, but your suggestion doesn't compile on my pc, even if you return -1 if T is complex, the rest of that function would still need to compile for complex values. It's possible to do, but I think adding an overload is a lot cleaner.

weslleyspereira: Hi. Ok, now I see the problem. The first routine won't work if T is complex. I am sorry for not noticing that. In this case, I agree there should be two versions: one for real types and other for complex types.

One comment: It is preferable to use `complex_type<T>` than `std::complex<T>` on the function signature. If nothing is said, `complex_type<T>` will be equivalent to `std::complex<T>`. There is the possibility, however, to use different structures for complex. The code is possibly not ready for that.

weslleyspereira (approved):

PR #38, opened 2022-04-01
add some assertions to matrix access

### LINKED ISSUE #37: Use of assertions
I think it's a good idea to use assertions in the vector and matrix access functions, for example:

```c++
        inline constexpr T&
        operator[]( idx_t i ) const noexcept {
            assert( i >= 0);
            assert( i < n);
            return (direction == Direction::Forward)
                ? *(ptr + (i*inc))
                : *(ptr + ((n-1)-i)*inc);
        }
```
I would add this to functions like `[]` and `()`, but also to `subvector`, `submatrix`, ....

This would massively help us to find bugs, and assertions are disabled during a production build so it won't actually cost us any performance.

thijssteel:
closes #37 

Adds a few assertions to the matrix access

### DIFF
--- a/include/legacy_api/legacyArray.hpp
+++ b/include/legacy_api/legacyArray.hpp
@@ -7,6 +7,8 @@
 #ifndef __TLAPACK_LEGACY_ARRAY_HH__
 #define __TLAPACK_LEGACY_ARRAY_HH__
 
+#include <assert.h>
+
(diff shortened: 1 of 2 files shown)


weslleyspereira:
Thanks a lot for that! Indeed, I also think the assert is a good command to use before access and slicing.

PR #39, opened 2022-04-01
Fix #35 ( and accidentally also fixed #31)

### LINKED ISSUE #31: Out of bounds in geqr2
I think i found a potential issue in geqr2. In the last iteration, `i` will be `k-1` and the start index of `x` is out of bounds. This isn't technically an issue because  `x` is also of size `0` and the value is never used, but with bounds checks enabled, this can be a problem.

https://github.com/tlapack/tlapack/blob/e99fdad4e9cad6573807b0624829f05899b8410e/include/lapack/geqr2.hpp#L82

### LINKED ISSUE #35: Fix LARF so that we do not have to edit A(i,i) to be one
I was reviewing @thijssteel PR#34 file gehd2.hpp, and I want to make a comment about larf.hpp. I have the same comment for geqr2.hpp. See: https://github.com/tlapack/tlapack/blob/a81590bada5b5424147175d24e1473f40f94771e/include/lapack/geqr2.hpp#L88

We should edit larf.hpp so that larf.hpp does not access A(i,i). 

A(i,i) in geqr2.hpp is v[0] in larf.hpp. 

v[0] should be one and we should not access it. We should assume that v[0] is one and go with it. 

The writing of v[0] = one in the routine that calls is an issue makes other parent routines not thread friendly. This happens more in larfb.hpp, ormqr.hpp, etc.

Let us fix this in larf.hpp to start with and then edit gehd2.hpp and geqr2.hpp accordingly. That would be a good start.

I understand that it is better to call ( one gemv.hpp and one ger.hpp ) instead of ( one gemv.hpp and one ger.hpp and one axpy.hpp, I think ), but the improved functionality is worth it in my opinion. 

The code would be cleaner and this is important for thread friendliness. 

One example of a related issue is: https://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=2&t=867

thijssteel:
Closes #35 and #31 

Fixing Issue #35 is a bit more difficult than I thought.

I could think of the following 

- make a copy of `v` inside `larf`, requiring some extra workspace and set `v[0]` to 1
- instead of calling `gemv` and `ger`, we do the calculation using loops
- we call `gemv` and `ger' on submatrices and then add the parts we skipped later
- use a specialized vector class that returns 1.0 for index 0 and v[i] otherwise

This PR is an attempt at the last option.

I can't get the type trait to work, but otherwise, it should be finished.

### DIFF
--- a/include/lapack/larfg.hpp
+++ b/include/lapack/larfg.hpp
@@ -109,6 +109,43 @@ void larfg( alpha_t& alpha, vectorX_t& x, tau_t& tau )
     }
 }
 
+/** Generates a elementary Householder reflection.
+ *
+ * larfg generates a elementary Householder reflection H of order n, such that
+ * 
+ *        H * ( alpha ) = ( beta ),   H' * H = I.
+ *            (   x   )   (   0  )
+ * 
+ * where alpha and beta are scalars, with beta real, and x is an (n-1)-element vector.
+ * H is represented in the form
+ * 
+ *        H = I - tau * ( 1 ) * ( 1 v' ) 
+ *                      ( v )
+ * 
+ * where tau is a scalar and v is a (n-1)-element vector.
+ * Note that H is symmetric but not hermitian.
+ * 
+ * If the elements of x are all zero and alpha is real, then tau = 0
+ * and H is taken to be the identity matrix.
+ * 
+ * Otherwise  1 <= real(tau) <= 2 and abs(tau-1) <= 1.
+ * 
+ * @param[in,out] v Array of length n.  On entry, the vector (alpha, x).  On exit, it is overwritten with the vector (beta, v).
+ * @param[out] tau On exit, the value tau.
+ * 
+ * @ingroup auxiliary
+ */
+template< class vector_t, class tau_t >
+void larfg( vector_t& v, tau_t& tau )
+{
+    using idx_t = size_type< vector_t >;
+    using pair  = std::pair<idx_t,idx_t>;
+
+    const idx_t n = size(v);
+    auto x = subvector( v, n > 1 ? pair{1,n} : pair{0,0} );
+    larfg( v[0], x, tau );
+}
+
 }
 
 #endif // __LARFG_HH__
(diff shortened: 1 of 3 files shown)


thijssteel:
I actually don't really like this solution. It's a complex solution just to set one element to 1, it might mess with some optimizations, ...
I'm just going to use loops.

thijssteel:
I actually don't really like this solution. It's a complex solution just to set one element to 1, it might mess with some optimizations, ...
I'm just going to use loops.

langou:
@thijssteel : I like option 3 most, I think. Option 2 might be good as well. 

Advantage of option 3 is that we can imagine that the data is off-loaded to a GPU so calling GEMV and GER would leave data on GPU and call these kernels on GPU, whereas for option 2, I think the whole data will have to be move from GPU to CPU. We can always write some specialized kernels. 

I think, for now, we should go with option 3.

langou:
"I actually don't really like this solution. It's a complex solution just to set one element to 1", well, it was fun to read. It's a good C++ exercise! I agree that this is not my favorite solution either. But it's fun to see what we can do with C++.

include/lapack/larfg.hpp
```diff
@@ -109,6 +109,43 @@ void larfg( alpha_t& alpha, vectorX_t& x, tau_t& tau )
...
+ * If the elements of x are all zero and alpha is real, then tau = 0
+ * and H is taken to be the identity matrix.
+ * 
+ * Otherwise  1 <= real(tau) <= 2 and abs(tau-1) <= 1.
+ * 
+ * @param[in,out] v Array of length n.  On entry, the vector (alpha, x).  On exit, it is overwritten with the vector (beta, v).
+ * @param[out] tau On exit, the value tau.
+ * 
+ * @ingroup auxiliary
+ */
```
weslleyspereira, line 137: I see a inconsistency with the documentation: `v` is used for 2 things. I would use `y` for instance on lines 122 and 123. The documentation is wrong, and #29 is there to fix it.

weslleyspereira (approved):
It looks good! The conflicts are easy to solve.

weslleyspereira:
I will deactivate the tests for `larf` from LAPACK++. On LAPACK, `larf` do not assume `v[0]=1`. On LAPACK++, the tester hits `larf` with vectors that do not begin with 1.

PR #54, opened 2022-04-19
Closes #19

### LINKED ISSUE #19: Missing extensive use of `tlapack_expr_with_vector`, `tlapack_expr_with_2vectors`, `tlapack_expr_with_vector_positiveInc`
While writing the performance example, I noticed the natural optimization associated incx=1 was not being used. The helpers: `tlapack_expr_with_vector`, `tlapack_expr_with_2vectors`, `tlapack_expr_with_vector_positiveInc` can help with that.

With these wrappers (maybe a couple more), it is easier to write the C++ wrappers in legacy_api in what concerns creating a vector directly from the raw pointer. Moreover, these helpers treat the case 'unitary increment' as a special case when creating the legacy vector from the raw pointer.

Here is how one of these wrappers is written:

```C++
    #define tlapack_expr_with_vector( x, TX, n, X, incx, expr ) do { \
        using internal::vector; \
        using internal::backward_vector; \
        if( incx == 1 ) { \
            auto x = vector( (TX*) X, n ); \
            expr; \
        } \
        else if( incx == -1 ) { \
            auto x = backward_vector( (TX*) X, n ); \
            expr; \
        } \
        else if( incx > 1 ) { \
            auto x = vector( (TX*) X, n, incx ); \
            expr; \
        } \
        else { \
            auto x = backward_vector( (TX*) X, n, -incx ); \
            expr; \
        } \
    } while(false)
```

A few routines (`asum`, `her`, `axpy`, `gemv`) use these helpers. All legacy_api wrappers using vectors can use these wrappers.

weslleyspereira:
Closes #19

This PR:
1. hides the complexity of negative increment and unitary increment to [`utils.hpp`](https://github.com/tlapack/tlapack/blob/c48d1a872ce129111455b4889a901bc849dd6089/include/legacy_api/blas/utils.hpp)
2. allows for the optimization with unitary increment in the same way as in Reference BLAS.
3. allows for negative increments on level 2 and level 3 BLAS routines on \<T\>LAPACK, e.g., hemm and her.

This PR also:
1. Removes quick returns that prevented inf and nan propagation from the wrappers.

### DIFF
--- a/include/legacy_api/blas/asum.hpp
+++ b/include/legacy_api/blas/asum.hpp
@@ -14,11 +14,11 @@
 namespace blas {
 
 /**
+ * Wrapper to asum( vector_t const& x ).
+ * 
  * @return 1-norm of vector,
  *     $|| Re(x) ||_1 + || Im(x) ||_1
  *         = \sum_{i=0}^{n-1} |Re(x_i)| + |Im(x_i)|$.
- *
- * Generic implementation for arbitrary data types.
  *
  * @param[in] n
  *     Number of elements in x. n >= 0.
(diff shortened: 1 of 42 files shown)


weslleyspereira:
The description was updated.

PR #59, opened 2022-04-29
Merges namespaces `lapack` and `blas` into the new `tlapack`

### LINKED ISSUE #23: Should we create `namespace tlapack`?
I think so. Having a single global namespace avoids repeating definitions. Moreover, it would make more sense in terms of usage.

Currently, <T>LAPACK has two namespaces: `blas` and `lapack`. This was inherited from the SLATE project. We could still maintain those namespaces behind `namespace tlapack`.

weslleyspereira:
Closes #23, and is related to #58.

This major change took me to the other minor changes:

- Homogenize the name of the macros limiting the scope of headers:
  - They will all start with `__TLAPACK_`.
  - `TBLAS` will only reference the header `tblas.hpp`.
- Avoid using `tlapack::` when the scope is clear.
- Create the `base` directory, that contain utility headers like: `arrayTraits.hpp`, `constants.hpp`, `types.hpp` and `utils.hpp`. The directories `blas` and `lapack` now only contain math routines.
- Change `BLAS_SIZE_T` to `TLAPACK_SIZE_T`, and `BLAS_INT_T` to `TLAPACK_INT_T`.
- Separates mpreal compatibility into a plugin.
- Use the contents of `config_lapackppwrappers.h` directly in `legacy_api/lapack.hpp`


thijssteel:
Looks good, i'd say, fix the failing CI builds and merge it as fast as possible so we don't have too much conflicts.

PR #62, opened 2022-05-02
Improve tests

thijssteel:
This PR adds a new test file for the Hessenberg reduction (and also fixes a few bugs i found while writing the tests)

It is a proposal for a pattern we can use for all tests. Several things I like:

- Template test cases to test 4 types with the same code
- Parameters are generated at the top of the file and reported in the dynamic sections. If a specific case fails, CATCH2 will then report the exact parameters used. You can then comment out the generator and replace it with a fixed value to run a single case. I have also made an issue in CATCH2 to maybe support only running a single dynamic section with a command line argument.

### DIFF
--- a/include/lapack/gehrd.hpp
+++ b/include/lapack/gehrd.hpp
@@ -137,10 +137,8 @@ namespace tlapack
         }
 
 
-        std::unique_ptr<TA[]> _Y(new TA[nb * n]);
-        std::unique_ptr<TA[]> _T(new TA[nb * nb]);
-        auto Y = colmajor_matrix<TA>(&_Y[0], n, nb, n);
-        auto T = colmajor_matrix<TA>(&_T[0], nb, nb, nb);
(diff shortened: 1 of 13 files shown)


langou (commented):
Good job!

thijssteel:
I'm quite happy with how the tests look, at least for now.

Extracting the matrix generation to a different class/function is probably going to be needed in the future, but i'll try to do that when we have a few more tests.

weslleyspereira (approved):
Looks good!!

Some comments:
- I think we should have the directory `test/include` to place `testutils.hpp`. Maybe also have a `test/include/defines.hpp` where the user could add/removes types from the tests. I had it in mind when proposed this: https://github.com/tlapack/testBLAS/tree/master/include
- Please, use something else than `[utils]` on the test tag, e.g. :

https://github.com/tlapack/tlapack/blob/711024c29f219dda2a8f97722f8757b59be131ca/test/src/test_lasy2.cpp#L20

PR #68, opened 2022-05-10
Add wrappers to all lv-3 BLAS

weslleyspereira:
This PR makes all lv-3 BLAS routines enable optimization via optmized BLAS (using BLAS++). The file 

https://github.com/weslleyspereira/tlapack/blob/try-all-lv3-blas-with-optmizedWrappers/include/legacy_api/blas/wrappers.hpp

has the lv-3 BLAS wrappers on \<T\>LAPACK.

### DIFF
--- a/test/src/test_gehrd.cpp
+++ b/test/src/test_gehrd.cpp
@@ -129,7 +129,8 @@ TEMPLATE_LIST_TEST_CASE("Hessenberg reduction is backward stable", "[eigenvalues
     DYNAMIC_SECTION("GEHRD with"
                     << " matrix = " << matrix_type << " n = " << n << " ilo = " << ilo << " ihi = " << ihi << " nb = " << nb)
     {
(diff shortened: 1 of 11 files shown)


test/src/test_gehrd.cpp
```diff
@@ -129,7 +129,8 @@ TEMPLATE_LIST_TEST_CASE("Hessenberg reduction is backward stable", "[eigenvalues
     DYNAMIC_SECTION("GEHRD with"
                     << " matrix = " << matrix_type << " n = " << n << " ilo = " << ilo << " ihi = " << ihi << " nb = " << nb)
     {
-        gehrd_opts_t<idx_t, T> opts = {.nb = nb, .nx_switch = 2};
+        gehrd_opts_t<idx_t, T> opts;
```
weslleyspereira, line 132: This is necessary because `gehrd_opts_t<idx_t, T> opts = {.nb = nb, .nx_switch = 2};` is only available on C++20.

PR #74, opened 2022-05-16
Remove (I believe) all identifiers starting with underscore + Uppercase

### LINKED ISSUE #73: Identifiers beginning with an underscore followed immediately by an uppercase letter are Reserved by the C++ standard
I just noticed this rule. It is written here for example: https://en.cppreference.com/w/cpp/language/identifiers
We should remove those identifiers from the library.

weslleyspereira:
This PR:

- Remove (I believe) all identifiers starting with underscore + Uppercase.

- Also, add the correct path when building `version.cpp`. It was failing on one version of the MinGW C++ compilers. [Discovered during the REU with students].

Closes #73


weslleyspereira:
This PR highlighted a bug for the Schur swap on Windows. The changes introduced here have nothing to do with the actual bug as far as I see, so I am merging it.

weslleyspereira:
This PR highlighted a bug for the Schur swap on Windows. The changes introduced here have nothing to do with the actual bug as far as I see, so I am merging it.

weslleyspereira:
This PR highlighted a bug for the Schur swap on Windows. The changes introduced here have nothing to do with the actual bug as far as I see, so I am merging it.

PR #87, opened 2022-05-26
check for beta == o in gemm to avoid nan propagation

thijssteel:
Jim mentioned in the last ballistic call that this is the intented behavior of gemm.


weslleyspereira (approved):
Looks good to me!

PR #92, opened 2022-05-31
initial draft of lu_mult

lindsay-slager:


include/lapack/lu_mult.hpp
```diff
@@ -0,0 +1,105 @@
...
+void lu_mult(matrix_t &A, const lu_mult_opts_t<size_type<matrix_t>> &opts = {})
+{
+    using idx_t = size_type< matrix_t >;;
+    using T = type_t<matrix_t>;
+    using range = std::pair<idx_t, idx_t>;
+    using real_t = real_type<T>;
+
+    const idx_t m = nrows(A);
+    const idx_t n = ncols(A);
+
```
thijssteel, line 50: You should assert that `nrows(A) == ncols(A)` and than `opts.nx >= 1`

include/lapack/lu_mult.hpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    const idx_t n0 = n / 2;
+
+    auto A00 = slice(A, range(0, n0), range(0, n0));
+    auto A01 = slice(A, range(0, n0), range(n0, n));
+    auto A10 = slice(A, range(n0, n), range(0, n0));
+    auto A11 = slice(A, range(n0, n), range(n0, n));
+
+    lu_mult(A11, opts);
```
thijssteel, line 84: I think your comments can be better here. Step 2, Step3, ... doens't say much if the reader doesn't have Julien piece of paper next to them. I would do something as follows:

- Describe how L and U are stored inside of A. Define L11 as the lower triangular part of A11, ...

- Describe four steps as follows:

1. A11 := L10*U01 + L11*U11
2. A01 := L00*U01
3. ...

test/src/test_lu_mult.cpp
```diff
@@ -0,0 +1,66 @@
...
+//
+// Copyright (c) 2022, University of Colorado Denver. All rights reserved.
+//
+// This file is part of <T>LAPACK.
+// <T>LAPACK is free software: you can redistribute it and/or modify it under
+// the terms of the BSD 3-Clause license. See the accompanying LICENSE file.
+
+#include <catch2/catch.hpp>
+#include <plugins/tlapack_stdvector.hpp>
+#include <tlapack.hpp>
```
thijssteel, line 12: you should also include `<plugins/tlapack_legacyArray.hpp>`

Your code currently works because `tlapack.hpp` also happens to include, but you may get in trouble if that changes in the future.

test/src/test_lu_mult.cpp
```diff
@@ -0,0 +1,66 @@
...
+    srand(1);
+
+    using matrix_t = TestType;
+    using T = type_t<matrix_t>;
+    using idx_t = size_type<matrix_t>;
+    typedef real_type<T> real_t;
+
+    idx_t n;
+
+    n = GENERATE(1, 2, 6, 9, 20, 23, 28);
```
thijssteel, line 29: 28 is a bit large for the unit tests, unless you have a good reason to test large matrices, i would limit it to 20. Remember that we are eventually testing this for thousands of routines for multiple different types and orientations and parameters. That can add up.

test/src/test_lu_mult.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    DYNAMIC_SECTION("n = " << n)
+    {
+        lu_mult(A);
+
+        // Calculate residual
+
+        gemm(Op::NoTrans, Op::NoTrans, real_t(1), L, U, real_t(-1), A);
+
+        real_t lu_mult_res_norm = lange(max_norm, A) / norma;;
```
thijssteel, line 63: double ;;

test/src/test_lu_mult.cpp
```diff
@@ -0,0 +1,66 @@
...
+    laset(Uplo::Upper, real_t(0), real_t(1), L);
+
+    laset(Uplo::Lower, real_t(0), real_t(0), U);
+    lacpy(Uplo::Upper, A, U);
+
+    real_t norma = lange(max_norm, A);
+
+    DYNAMIC_SECTION("n = " << n)
+    {
+        lu_mult(A);
```
thijssteel, line 57: you should try different values for nx.

thijssteel (changes requested):

include/lapack/lu_mult.hpp
weslleyspereira: Solved!

include/lapack/lu_mult.hpp
weslleyspereira: Done! Thanks for the suggestions!

test/src/test_lu_mult.cpp
weslleyspereira: Done!

test/src/test_lu_mult.cpp
weslleyspereira: Thanks!

thijssteel (approved):

PR #94, opened 2022-05-31
Tester for GEBD2

yuxincai11:
So far the tester for GEBD2 is finished and passed the tests. It uses GEBD2 (to generate reflectors and an **upper bidiagonal** matrix B), UNG2R (to do QR and gives lower Q), and UNGL2 (to do LQ and gives upper Z). Also included three checks: two ortho_check for unitary matrices Q and Z and one final test for B = Q_H * A * Z (while in real time, the Z given by UNGL2 is actually Z_H), so in the end we are checking B = Q_H * A * Z_H.

Note: this gebd2 only supports m >= n, that's why the B it produces is only an upper bidiagonal.
Later when the lower bidiagonal is included, the condition would be for upper bidiagonal (m >= n); lower bidiagonal (m < n).


include/lapack/gebd2.hpp
```diff
@@ -0,0 +1,65 @@
+#ifndef GEBD2
```
thijssteel, line 1: have a look at other <T>LAPACK codes, they all have comments at the start. Please do the same.

include/lapack/gebd2.hpp
```diff
@@ -0,0 +1,65 @@
...
+#define GEBD2
+
+// #include <utility>
```
thijssteel, line 4: Remove comments

include/lapack/gebd2.hpp
```diff
@@ -0,0 +1,65 @@
...
+// #include <plugins/tlapack_debugutils.hpp>
+// #include <tlapack.hpp>
+
+#include "base/utils.hpp"
+#include "base/types.hpp"
+#include "lapack/larfg.hpp"
+#include "lapack/larf.hpp"
+
+/** 
+* hand-crafted  reduction to bidiagonal form of a general m-by-n matrix A, level 2, right looking algorithm
```
thijssteel, line 16: These comments should be above your function definition

They are also not complete yet. Have a look at other <T>LAPACK function for examples of detailed documentation.
You can also copy documentation from [LAPACK](https://www.netlib.org/lapack/explore-html/d3/d01/group__complex16_g_ecomputational_gad5a784fc93cbf57b6b8ff3f0ff229cbe.html)

include/lapack/gebd2.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+
+    const idx_t m = nrows( A );
+    const idx_t n = ncols( A );
+
+    for (idx_t j = 0; j < n; ++j) {
+
+        auto v = slice(A, range(j, m), j);
+        
+        larfg(v, tauv[j]); //gen the vertical reflector v's
```
thijssteel, line 38:
```suggestion
        larfg(v, tauv[j]); //genenerate the vertical reflector v
```

include/lapack/gebd2.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+        if( j < n-1 ){ 
+            auto A11 = slice(A, range(j, m), range(j+1, n));
+            larf(Side::Left, v, conj(tauv[j]), A11, work);
+        }
+
+        if( j < n-1 ){
+            auto w = slice(A, j, range(j+1, n)); 
+            //for loop to conj w.
```
thijssteel, line 47: This is not a valuable comment, anyone can see that the for loop is conjugating w. It would be much better to comment **why** you are conjugating it.

include/lapack/gebd2.hpp
```diff
@@ -0,0 +1,65 @@
...
+            for (idx_t i = 0; i < n-j-1; ++i)
+                w[i] = conj(w[i]);
+
+            larfg(w, tauw[j]); //gen the horizontal reflector w's
+
+            if( j < m-1 ){ 
+                auto B11 = slice(A, range(j+1, m), range(j+1, n));
+                larf(Side::Right, w, tauw[j], B11, work);
+            }
+            // for (idx_t i = 0; i < n-j-1; ++i) //for loop to conj w back.
```
thijssteel, line 57: remove commented code

thijssteel (changes requested):

include/lapack/gebd2.hpp
yuxincai11: Already updated the comments here. 

It's originally from LAPACK, but Weslley found out that we might not need this "conjugate back" part, along with a few other LAPACK conjugating "back-and-forth" when generating Z. Will check with Julien about this! Thanks!

thijssteel (approved):

weslleyspereira:
Nice job @yuxincai11! I think this PR is ready to merge.

PR #98, opened 2022-06-01
GELQ2 + UNGL2 (with tester)--Ready for review

yuxincai11:


langou:
Two comments:

(1) For an m-by-n matrix, UNG2L allows for k to be less than m. LAPACK does not allow this. We need to think whether this is something we want to keep or to remove. If user wants k to be less than m, then user can simply slice the matrix and give k rows (instead of m.) I think the answer is how much the <T>LAPACK code is polluted by enabling k to be less than m. If the code reads fine, we leave. If the code does not read well, we remove.

(2) There is also the argument "bool &bidg" in UNG2L which is about whether the Householder vectors are conjugated or not. The goal of GELQ2 is to return the factorization (W on top of L) for A so that (W on top of L) is conjugate transpose of (R on top of V) for GEQR2 of conjugate transpose of A. It can be useful to waive this property indeed. I would suggest to change the name of the variable. And to include the variable in GELQ2 as well.

These are more design decision. The code looks good overall.

yuxincai11:
> Two comments:
> 
> (1) For an m-by-n matrix, UNG2L allows for k to be less than m. LAPACK does not allow this. We need to think whether this is something we want to keep or to remove. If user wants k to be less than m, then user can simply slice the matrix and give k rows (instead of m.) I think the answer is how much the LAPACK code is polluted by enabling k to be less than m. If the code reads fine, we leave. If the code does not read well, we remove.
> 
> (2) There is also the argument "bool &bidg" in UNG2L which is about whether the Householder vectors are conjugated or not. The goal of GELQ2 is to return the factorization (W on top of L) for A so that (W on top of L) is conjugate transpose of (R on top of V) for GEQR2 of conjugate transpose of A. It can be useful to waive this property indeed. I would suggest to change the name of the variable. And to include the variable in GELQ2 as well.
> 
> These are more design decision. The code looks good overall.

include/lapack/gelq2.hpp
```diff
@@ -60,24 +61,37 @@ namespace tlapack
...
         const idx_t n = ncols(A);
         const idx_t k = std::min(m, n);
 
+        // check arguments
+        tlapack_check_false(access_denied(dense, write_policy(A)), -1);
+        tlapack_check_false((idx_t)size(tauw) < std::min<idx_t>(m, n), -2);
+        tlapack_check_false((idx_t)size(work) < m, -3);
+
+        // quick return
+        if (m <= 0)
```
yuxincai11, line 75: @weslleyspereira Hey Weslley, I referred to your geqr2.hpp here and changed the quick return index `n` to `m` since the `work` for gelq2 is always `m`. Is this okay here?

include/lapack/gelq2.hpp
```diff
@@ -60,24 +61,37 @@ namespace tlapack
...
+            larfg(w, tauw[j]);
 
             if (j < k - 1 || k < m)
             {
+                // Apply H(j) to A(j+1:m,j:n) from the right
                 auto Q11 = slice(A, range(j + 1, m), range(j, n));
                 larf(Side::Right, w, tauw[j], Q11, work);
             }
-            for (idx_t i = 0; i < n - j; ++i) 
+            for (idx_t i = 0; i < n - j; ++i)
```
yuxincai11, line 94: @langou Hey Julien, so the algorithms for GEBD2 and GELQ2 are similar here, they conjugate `w` before and after `larfg` and `larf`. In GELQ2 (and the UNGL2 `else` part to pair with it) I just used the LAPACK ZGELQ2 here. 

So Weslley found that we can just conjugate `w` for once before `larfg` in GEBD2 (and no conj(w) in UNGL2 `if (bidg == 1) ` part, but with `conj(tauw)` in `larf),` and this is pretty much the same for GELQ2 (in this line 94 `for` loop can go away along with the two `conj(w)`s in UNGL2). 

You mentioned you want me to put the bidiagonal option in GELQ2 but we don't really use it here. Do you still want me to include the bidiagonal in it? Would you want me to change the `conj(w)` algorithms for GELQ2 and UNGL2 (as GEBD2) as well?

yuxincai11: I decided that I'll go two different routines for UNGL2 and generating matrix Z for GEBD2, since they are indeed doing different things. Therefore, I'll do another .hpp file for generating Z and remove the bidiagonal boolean in UNGL2 to make it cleaner.

weslleyspereira: I think @thijssteel and I agree with splitting into two routines. Nice job @yuxincai11!

include/lapack/gelq2.hpp
```diff
@@ -0,0 +1,99 @@
...
+            }
+            for (idx_t i = 0; i < n - j; ++i)
+                w[i] = conj(w[i]); 
+        }
+
+        return 0;
+    }
+}
+
+#endif // __TLAPACK_GELQ2_HH__
```
weslleyspereira, line 99: FYI - This red symbol "(-)" means: No newline at end of file. This is a requirement of old versions of C++. We don't need to put a new line at the end of the file. See:

"A source file that is not empty and that does not end in a new-line character, or that ends in a new-line character immediately preceded by a backslash character before any such splicing takes place, shall be processed as if an additional new-line character were appended to the file (C++11 §2.2/1)." (https://stackoverflow.com/a/8172348/5253097)

I couldn't find the proper reference because I don't dispose of a complete C++11 manual.

include/lapack/gelq2.hpp
```diff
@@ -0,0 +1,99 @@
...
+//
+// This file is part of <T>LAPACK.
+// <T>LAPACK is free software: you can redistribute it and/or modify it under
+// the terms of the BSD 3-Clause license. See the accompanying LICENSE file.
+
+#ifndef __TLAPACK_GELQ2_HH__
+#define __TLAPACK_GELQ2_HH__
+
+#include <plugins/tlapack_stdvector.hpp>
+#include <plugins/tlapack_legacyArray.hpp>
```
weslleyspereira, line 15: You may remove those lines. Those plugins are useful when you use `std::vector` and `tlapack::legacyMatrix` in your routine.

PR #106, opened 2022-06-15
adding tester for gelqf

yuxincai11:


weslleyspereira:
In my machine, your tests fail to pass some bounds checks. I think the following changes make sense:

```diff
diff --git a/include/lapack/gelqf.hpp b/include/lapack/gelqf.hpp
index c576316..1afe1dd 100644
--- a/include/lapack/gelqf.hpp
+++ b/include/lapack/gelqf.hpp
@@ -68,13 +68,14 @@ namespace tlapack
         // constants
         const idx_t m = nrows(A);
         const idx_t n = ncols(A);
+        const idx_t k = std::min<idx_t>(m, n);
 
         // check arguments
         tlapack_check_false(access_denied(dense, write_policy(A)), -1);
-        tlapack_check_false((idx_t)size(tauw) < std::min<idx_t>(m, n), -2);
+        tlapack_check_false((idx_t)size(tauw) < k, -2);
         tlapack_check_false((idx_t)size(work) < m, -3);
 
-        for (idx_t j = 0; j < m; j += nb)
+        for (idx_t j = 0; j < k; j += nb)
         {
             // Use blocked code initially
             idx_t ib = std::min<idx_t>(nb, m - j);
@@ -86,7 +87,7 @@ namespace tlapack
 
             gelq2(A11, tauw1, work);
 
-            if (j + ib < m)
+            if (j + ib < k)
             {
                 // Form the triangular factor of the block reflector H = H(j) H(j+1) . . . H(j+ib-1)
                 larft(Direction::Forward, StoreV::Rowwise, A11, tauw1, TT1);
```

weslleyspereira:
Hi! I have just rebased your branch, @yuxincai11. I have a couple of contributions to give to this thread. I will comment on this Conversation thread as soon as I submit the modifications. I will try to be very organized so you can follow or contribute if you want. Thanks again for the job!

yuxincai11:
Hey Weslley @weslleyspereira, 

I think the failing might be because of the deletion of those for loops to conjugate `w` in ungl2. When I was testing it on my end, if we want ungl2 to work for both gebd2 and gelq2 (and gelqf), we must keep the conjugates even though in some events we can skip them (and the skipping must be in pair, like in both gebd2 and ungl2). That's why Julien and I decided to keep the conjugations in ungl2 eventually. Just my guess on the conj(w) cuz I bumped into a bunch of problems with them back then in both gebd2 and gelq2, but it could also be something else.

weslleyspereira:
Hi @yuxincai11. I didn't expressed clearly, sorry. There were no tests failing; the problem was an incompatibility with the new version of mdspan. I am proposing a PR to solve it: https://github.com/tlapack/tlapack/pull/108.

Also, I kept the conjugates as you proposed. I have just put them inside the loop, so we avoid it when they aren't necessary.

weslleyspereira (approved):

yuxincai11:
I see! I didn't notice the changes below since there were a lot of them :) The codes look very much better now. Thanks, @weslleyspereira !

PR #123, opened 2022-07-30
Fix gemm legacy interface and propose a gemm with no beta

weslleyspereira:
- gemm is the interface from BLAS. Since the interface in T-LAPACK is different, it makes sense to rename it.
- Also, by renaming it we could make different design choices.

`matrix_multiply` has two versions: one with and one without beta.
1. The first version computes `C := alpha*A*B + beta*C`;
2. The second computes `C := alpha*A*B`.
They are semantically different: in the first, C is [in,out]; in the second, C is [out] only.
Legacy `gemm` covers both functionalities depending on the value of beta (if beta is zero, use second expression).
The problem with the strategy adopted in `gemm` is: if beta evaluates to zero in some part of an algorithm and there are Infs or NaNs in C, those will not propagate.

The proposal is: if one wants to obtain the same behavior of gemm and
- `alpha=0` and `beta=0`: just use `laset` to set the matrix C to 0.
- `alpha=0` and `beta` is not zero: just use `lascl` to scale the matrix C.
- `alpha` is not zero and `beta=0`: use the version of `matrix_multiply` that considers C as output only.
- otherwise, use `matrix_multiply` that considers C as input and output.

This PR also introduces the StrongZero internal data type. This is to avoid code duplication.
Also, it solves failing tests in https://github.com/tlapack/testBLAS/runs/7583337777.

### DIFF
--- a/examples/eigenvalues/example_eigenvalues.cpp
+++ b/examples/eigenvalues/example_eigenvalues.cpp
@@ -200,8 +200,8 @@ void run(size_t n)
             for (size_t i = 0; i < n; ++i)
                 work(i, j) = static_cast<float>(0xABADBABC);
 
-        tlapack::gemm(tlapack::Op::NoTrans, tlapack::Op::NoTrans, (T)1.0, Q, H, (T)0.0, work);
-        tlapack::gemm(tlapack::Op::NoTrans, tlapack::Op::ConjTrans, (T)1.0, work, Q, (T)0.0, H);
+        tlapack::gemm(tlapack::Op::NoTrans, tlapack::Op::NoTrans, (T)1.0, Q, H, work);
+        tlapack::gemm(tlapack::Op::NoTrans, tlapack::Op::ConjTrans, (T)1.0, work, Q, H);
 
         for (size_t j = 0; j < n; ++j)
             for (size_t i = 0; i < n; ++i)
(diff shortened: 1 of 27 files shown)


weslleyspereira:
After talking to @langou, I decided to address only the following part on this PR:

Split the 2 behaviors of gemm (and other matrix multiplication routines) as follows:
1. The first version computes `C := alpha*A*B + beta*C`;
2. The second computes `C := alpha*A*B`.
They are semantically different: in the first, C is [in,out]; in the second, C is [out] only.
Legacy `gemm` covers both functionalities depending on the value of beta (if beta is zero, use second expression).
The problem with the strategy adopted in legacy `gemm` is: if beta evaluates to zero in some part of an algorithm and there are Infs or NaNs in C, those will not propagate. Those 2 versions aims to improve it.

If one wants to obtain the same behavior of legacy `gemm` and
- `alpha=0` and `beta=0`: just use `laset` to set the matrix C to 0.
- `alpha=0` and `beta` is not zero: just use `lascl` to scale the matrix C.
- `alpha` is not zero and `beta=0`: use the version of `gemm` that considers C as output only.
- otherwise, use `gemm` that considers C as input and output.

This commit also introduces the StrongZero internal data type. This is to avoid code duplication without losing performance. StrongZero evaluates to `0` at compile time.

TO-DO list:
- [ ] Apply similar modifications to all lv-3 BLAS routines.
- [ ] Change the documentation.
- [ ] Add tests to the test suite.

weslleyspereira:
About the internal class `StrongZero`:

This is class, for now, defines multiplication and sum with another class T. Other functionality may be added in future if needed.

Suppose x is of type T. Then: 

1. T(StrongZero()) is equivalent to T(0). 
2. x *= StrongZero() is equivalent to x = T(0). 
3. x += StrongZero() does not modify x. 
 
This class satisfies: 
 
     x * StrongZero() == StrongZero() 
     StrongZero() * x == StrongZero() 
     x + StrongZero() == x 
     StrongZero() + x == x  

https://github.com/tlapack/tlapack/blob/f84114d665f6a0fce336cc0b0f903b35decdcf67/include/tlapack/base/types.hpp#L60-L110

I use this class inside the wrapper for `C := alpha*A*B` as `return gemm( transA, transB, alpha, A, B, internal::StrongZero(), C )`. See:

https://github.com/tlapack/tlapack/blob/f84114d665f6a0fce336cc0b0f903b35decdcf67/include/tlapack/blas/gemm.hpp#L200-L254

PR #128, opened 2022-08-24
Try new workspace design

weslleyspereira:
I came back to the discussion #44 and would like to share my proposal. This was based on the discussions with @thijssteel, @langou and @mgates3.

# Considering the implementation of workspaces and workspace queries

Some routines need extra space that depends on the size of the input matrices and vectors. LAPACK calls it workspace.

## Optional vs Required workspace

A fundamental design question is: should the workspace be allocated dynamically inside a routine or provided from outside? Here is what some packages for Numerical Linear Algebra do:
- LAPACK requires workspace to be passed through the interface of the routine. So, no dynamic memory allocation inside the routine.
- LAPACKE provides two interfaces: one requires workspace as in LAPACK, and the other does not. The latter allocates workspace dynamically inside the routine.
- SLATE, LAPACK++ and Eigen allocate workspace dynamically inside their routines.

In \<T\>LAPACK, workspace is optional. The caller can either define the workspace outside the routine and inform it to its interface, or pass a zero-sized workspace. The latter automatically triggers an internal dynamic memory allocation. If insufficient size is provided, we return an error to avoid confusion.

## Workspace type

In LAPACK, each workspace is an array of the required type, e.g., single precision, double precision, complex and integer. Routines may need multiple workspaces because of that. Since LAPACKE is a wrapper to LAPACK, it implements the same type dependency logic.

In \<T\>LAPACK, workspace is a structure that contains an array of bytes (`unsigned char *`) and its size (`std::size_t`). When needed, we allocate bytes using an allocator that defaults to `std::allocator<unsigned char>`. Routines that need workspace provide the option `work_t` which is the matrix type that use the workspace. Since this an option, not a requirement, there is always a default matrix type if none is provided. In general, routines use an instance of the template `legacyMatrix` as the default value for `work_t`. Routines that need multiple workspaces must provide multiple matrix types, e.g., `work_t, rwork_t, iwork_t`. 

## Workspace query

Workspace queries give the optimal workspace a routine needs to execute. In LAPACK, there is only workspace queries for the first workspace of each routine, which is a design flaw. Moreover, the optimal workspace is returned using a non-integer type, which is not ideal and raise issues from time to time. Each routine that uses workspace runs a query mode additionally to its main functionality.

In \<T\>LAPACK, workspace queries are separate functions that return the optimal number of bytes. `size_t foo_worksize( ... )` is the signature of the workspace query of a routine `foo( ... )`. All routines that need workspace must provide the workspace query.

## Optimal and minimum workspace sizes

Optimal workspace size is the one which the routine is expected to have lower execution times. Minimum workspace size refers to the necessary space for that routine to work.

In \<T\>LAPACK, the value of the optimal workspace size may not present in the documentation of a routine, since this value may vary from implementation to implementation. We recommend the caller to always perform a workspace query to obtain the optimal workspace size. The minimum workspace size, if different from the optimal workspace size, must be provided with the documentation of a routine. A function will return an error code if the size of the provided workspace is lower than the minimum.

### Optional workspace

Default options for functions that have only a single workspace (https://github.com/weslleyspereira/tlapack/blob/ad2a1a8bfd02fcbc47e040f69d9d71be32f00c9d/include/tlapack/base/utils.hpp#L994-L1005)

```c++
/// Workspace options
template<
    class T,
    class idx_t = std::size_t,
    class work_type = legacyMatrix<T,idx_t>
>
struct workspace_opts_t
{
    using work_t = work_type;   ///< Workspace type

    byte* work = nullptr;       ///< Workspace
    std::size_t lwork = 0;      ///< Workspace size
};
```

`byte` is defined at https://github.com/weslleyspereira/tlapack/blob/ad2a1a8bfd02fcbc47e040f69d9d71be32f00c9d/include/tlapack/base/types.hpp#L338-L342
```c++
    /// Byte type
    using byte = unsigned char;
    /// Byte allocator
    using byteAlloc = std::allocator<byte>;
    /// Vector of bytes. May use a specialized allocator in future
    using vectorOfBytes = std::vector<byte,byteAlloc>;
```

and this is how we deal with optional workspaces inside a routine (example for `larfb`):

```c++
    // Allocates workspace
    vectorOfBytes localwork;
    size_t lwork = larfb_worksize( side, trans, direction, storeMode, V, Tmatrix, C, opts );
    const byte* work = alloc_workspace( localwork, lwork, opts.work, opts.lwork );
```

The function `alloc_workspace` will allocate `lwork` bytes in `localwork` unless optional workspace is provided.

Comments:

- It is important to have control of the byte pointer for the workspace instead of using a `std::vector`. This is because, inside our routines, we recurrently need to slice the workspace. We cannot do that with `std::vector` as far as I know.

- Having a default option type for workspace, i.e., `workspace_opts_t`, helps routines to pass options internally. See how `unmqr` calls `larfb` at https://github.com/weslleyspereira/tlapack/blob/ad2a1a8bfd02fcbc47e040f69d9d71be32f00c9d/include/tlapack/lapack/unmqr.hpp#L166: `larfb( side, trans, forward, columnwise_storage, V, Tmatrix, Ci, opts );`. This works because:
  1. `larfb` uses `workspace_opts_t` as options.
  2. The options for `unmqr` is derived from `workspace_opts_t`.

  This said, it is important to mention that it is not mandatory that the routines options derives from `workspace_opts_t`, but this should be used whenever we can take some advantage.
  
- C++20 implements `byte` as subclass of `unsigned char`. Currently, we do not need anything from that class, so we can use byte as just an alias to `unsigned char`.

- `byteAlloc` is an allocator. Depending on the machine / system, it can be helpful to have a allocator other than `std::allocator<byte>`. For instance, @mgates3 explained to me the design of LAPACK++ in a recent email:

> In LAPACK++ we use a custom C++ allocator to do 2 things for the workspace:
> 1) Don't construct the memory, i.e., don't zero it out. The overhead was causing issues for end users on Xeon Phi machines.
> 2) Align to 64 bytes (512 bits), which should make AVX-512, etc. happier.
> See https://bitbucket.org/icl/lapackpp/src/master/src/NoConstructAllocator.hh, with lapack::vector at bottom.

### Other changes:
- Legacy classes for matrix and vectors are part of the main library. They are no longer only used for the legacy interface wrappers. They are used to build matrices and vectors from workspace arrays. Because of that, their index type `idx_t` is now a template. Before, they were associated to the `idx_t` from the legacy interface wrappers. Also, there is no plugin for those classes anymore since they were moved to the inside of the library.
- There is a new item in the interface of data objects: `Create`. This is a C++ functor. It implements the way new matrices and vectors are created. Look at the documentation at arrayTraits.hpp . With this new functionality, the internal routines `tlapack::internal::colmajor_matrix` and `tlapack::internal::rowmajor_matrix` were removed from all examples and tests. Those are for internal usage only.

## TODO:

- [x] Apply new design to all routines.
- [x] Replace the use of `legacyMatrix` by the new way of creating matrices in the tests.
- [x] Update the plugins for `Eigen, kokkos::mdspan` so that they implement the `tlapack::Create` class.
- [ ] Update documentation.


mgates3:
Some thoughts:
In LAPACK++, we recently added a few routines with workspace queries for GPU functions. We named these:
```
geqrf_work_size_bytes
getrf_work_size_bytes
potrf_work_size_bytes
```
to emphasize that they are in bytes, not doubles or whatever (as LAPACK would do).
We also use output arguments, rather than returning the work space size, because we needed both host & GPU work space sizes. ("Needed" being a loose term; cuSolver has host sizes in its interface, but so far they are always 0.) The interface is:
```
void geqrf_work_size_bytes(
    int64_t m, int64_t n,
    scalar_t* dA, int64_t ldda,
    size_t* dev_work_size, size_t* host_work_size,
    lapack::Queue& queue )
```

You mention minimum & optimal workspace sizes. Is there any mechanism to query both? E.g., `dgeqr` (not `dgeqrf`) returns either one or the other, while `zgesvdq` returns both optimal and minimum in CWORK[1] and CWORK[2].

weslleyspereira:
Thanks @mgates3! I am still editing the proposal. Sorry if you received an email for the incomplete version. I will send a new message when I am done with the description.

> We also use output arguments, rather than returning the work space size, because we needed both host & GPU work space sizes. ("Needed" being a loose term; cuSolver has host sizes in its interface, but so far they are always 0.) The interface is:

I see. Thanks for letting us know. I think you have the better design. I will move the return value to an output arguments.

> You mention minimum & optimal workspace sizes. Is there any mechanism to query both? E.g., dgeqr (not dgeqrf) returns either one or the other, while zgesvdq returns both optimal and minimum in CWORK[1] and CWORK[2].

Since there are output arguments, we could think on returning the minimum workspace in another argument. In your case, something like:

```
void geqr_work_size_bytes(
    int64_t m, int64_t n,
    scalar_t* dA, int64_t ldda,
    size_t* dev_work_size, size_t* host_work_size,
    size_t* dev_work_min, size_t* host_work_min,
    lapack::Queue& queue )
```

Queries for the minimum space could also be implemented in another routine, unless there is a known issue with returning those workspaces using different functions.

weslleyspereira:
I finished editing the initial proposal in the first message, @mgates3, @thijssteel and @langou. I will now consider @mgates3's comments.

weslleyspereira:
The proposal should be ready for review. If the proposal is good, I will proceed with my TO-DO list.

I considered @mgates3's comments in [9ff828e](https://github.com/tlapack/tlapack/pull/128/commits/9ff828e85033ccb8ef97ee47ca5b23e3edcdd653). Thanks! Other changes:
- `alloc_workspace` now consider both optimal and minimum workspace sizes, as it should be. I also improved the exception handling inside it.
- `unmqr_worksize` now calls `larfb_worksize` to get optimal and minimum workspace sizes. I believe this is the kind of design  we want.

thijssteel:
Some thoughts:

- I'm not sure that minimal workspace is something we want to consider. Usually, the required workspace is not very large if you also consider the size of the matrices. And I've seen a few times where people supply the minimal workspace without fully realizing just how badly they are handicapping the routine. On top of that. It complicates our routines and requires more maintenance and testing (we need to make sure it actually works with that minimal workspace).

- I think it would be good to have a mechanism that checks whether the created matrix actually fits into the buffer you provided. This can be a simple assertion, but we do need to pass the size.

- I've previously had a discussion with Julien about the memory size of matrices. It seems that we assume here that an `m x n` matrix will always take up `m x n x nbytes(T)` bytes in memory. But maybe some matrices are constructed with a larger stride for efficiency, or they could be sparse or distributed. I agreed with Julien at the time, but now I think it may be too much complexity at this stage of the library. Let's get the dense stuff working first.

Otherwise it looks good to me. Good work, I like the functor design.

langou:
I am not convinced we need "minimum workspace" size too. 

I think it might be a good idea to have the tentative feature in place (feature = give user choice whether they want optimal workspace, minimum workspace, etc.) in case we decide to work on it in the future. Not sure we will ever do that though. But having the feature in place (a mechanism to query a various workspace sizes) could be good. 

I think a library software package that is working only with optimal workspace is not too bad.

It is not clear to me how many users are using a workspace size below ``optimal``. It seems a fraction. And the one who are doing it, as @steelt wrote, are often not ``fully realizing just how badly they are handicapping the routine`` in term of performance.

Also, the optimal workspace size is only a fraction of the memory needed to run the application. So it is not clear when using a workspace size below optimal is critical for an application. N is large-ish compare to NB. Matrices are dense and N^2. And workspace is N*NB. So if you cannot afford N*NB you are really tight in space. (Many assumptions here, N is not that much larger than NB in some applications, workspace can be N^2, etc.)

The reason we started to implement (once!) workspace query for minimal workspace size in LAPACK was because it was inconsistent with how optimal workspace is treated. It is kind of a backward compatibility problem for LAPACK when we write in the comments ``the code will run with a workspace of size N``. It does limit our possibilities to change things in the future. (We can always change and backup on the old code, but this is ugly and force us to maintain the old code.) So the idea was to never write workspace size (optimal, minimal, etc.) in comments and always have users querying it. The idea is to say to user ``do not assume a workspace size``. Query before using.

Finally, the whole design what if the user provide a workspace between ``minimal`` and ``optimal`` in LAPACK is ``complicated``. LAPACK tries to find a block size in as large as possible to attempt to provide as much as performance, but it is not clear the effort in the LAPACK code warrants the usefulness of the functionality. 

All in all:
(*) one workspace size (optimal) would be OK
(*) planning for various workspace sizes but not doing anything about it for now (only optimal) would be OK. Not clear we will ever go that road though

weslleyspereira:
Comments:
- Is there any known relevant benefits for using a minimum workspace? If the assumption "in almost all situations the caller will have enough space for workspace" is reasonable, I would rather just remove the words "minimum" and "optimal" from the design and just have a single "workspace". @thijssteel has a good point about maintenance and testing. There are already too many configurations to test.
- I agree that for backward compatibility we shouldn't mention the computed size of an workspace query on the documentation. I will remove that from the documentation of the queries as well.
- @thijssteel said `I think it would be good to have a mechanism that checks whether the created matrix actually fits into the buffer you provided. This can be a simple assertion, but we do need to pass the size.` -> Agreed! I was planning to have tests for that, but it is better to have error checks in place.

langou:
If the assumption "in almost all situations the caller will have enough space for workspace" is reasonable, I would rather just remove the words "minimum" and "optimal" from the design and just have a single "workspace". 
=> +1

weslleyspereira:
I think I finally got a good design. I will write about it in this thread, as well as update documentation and maybe do some cleaning

weslleyspereira:
# Current (I hope final) modifications introduced in this PR

## Workspace

A workspace is a column-major matrix of bytes. As such, it has the following attributes:
- Sizes `size_t m, n`, which represent "n pieces of a contiguous memory of size m bytes".
- Leading dimension `size_t ldim` in bytes.
- Pointer to a position in memory `unsigned char* ptr`. `unsigned char` is the type we use to measure bytes, as in https://en.cppreference.com/w/cpp/types/byte.

Therefore, we could have implemented the Workspace class as a subclass of `legacyMatrix` with data type `unsigned char`. We chose not to do so, to make the two classes independent of one another.

This design was meant to cover all possibilities of workspaces in LAPACK. In fact,
1. It allows for the use of both contiguous and non-contiguous arrays as workspace. Non-contiguous arrays are useful for some routines since they may use part of their matrices and vectors as workspace.
2. It covers row-major and column-major matrices, and vectors with various increments. This is because all those array types can be mapped to a column-major matrix. A vector with N double precision numbers with increment inc, for instance, can be mapped to a workspace with sizes `m = sizeof(double); n = N; ldim = inc*sizeof(double)`.
3. It allows for multiple workspace arrays in the interface of a routine, and this is because we count information in bytes. It is the responsibility of each routine to give shape to its workspace.

Particular choices:

1. The objects of this class always maintain one of the states:
   a. `(n > 1) && (m > 0)`.
   b. `((n <= 1) || (m == 0)) && (ldim == m)`.
This is mainly to ease the check for a contiguous workspace. Observe that we do not assume that `ldim > 0` when m is 0.

2. The function `Workspace Workspace::extract( size_t m, size_t n )` returns a workspace that is obtained by removing `m*n` bytes from the current one. If the starting workspace is not contiguous, then we require `this->m == m || this->n == n`. This is required to prevent bad partitioning by the algorithms using this functionality.

3. The implementation is separated in the file `include/tlapack/base/workspace.hpp`.

## Matrix and vector creation

Since the workspace is a very simple structure, we need some strategy to give them the shape of a specific matrix or vector type. Suppose we want to use the allocated workspace `work` as a `tlapack::legacyMatrix<float>`. In this case, we want to call a routine `Create< tlapack::legacyMatrix<float> >( work )` that will return the formatted object relative to the given type. This is exactly what we do here. `Create` is a special kind of structure called functor (not function!). See, e.g., [www.geeksforgeeks.org/functors-in-cpp](https://www.geeksforgeeks.org/functors-in-cpp/).

Every new data type `matrix_t` that will be used with \<T\>LAPACK needs to implement its own `CreateImpl` functor with a couple of functionalities:
1. `std::vector<T> v; Create<matrix_t>( v, m, n );` to create new objects without preallocated memory.
2. `Workspace w; Create<matrix_t>( w, m, n );` to create new objects from preallocated memory.
3. `Workspace wIn; Workspace wOut; Create<matrix_t>( wIn, m, n, wOut );` to create new objects from preallocated memory, and obtain the remaining workspace.

For vectors, the interface has only 1 entry for size instead of 2.

## Optional parameters:

Optional parameters are the last entry of a routine in \<T\>LAPACK. Not all routines have options. An option is a single object passed by reference. The object belongs to a structure that can have a variety of attributes like: workspace, error handling configurations, block sizes, etc. Some options have their own function attributes.

Options can be derived from other options, which is the case of all options that have a workspace. This design avoids code repetition and eases the cast from one option to another in internal calls. Two examples of default options are as below:

```c++
/// Error Checking options
struct ec_opts_t {
   ErrorCheck ec = {};

   inline constexpr
   ec_opts_t( const ErrorCheck& ec_ = {} ) : ec(ec_) {}
};

/**
 * @brief Options structure with a Workspace attribute
 * 
 * @tparam work_t Give specialized data type to the workspaces.
 *      Behavior defined by each implementation using this option.
 */
template< class ... work_t >
struct workspace_opts_t
{
   Workspace work; ///< Workspace object

   // Constructors:

   inline constexpr
   workspace_opts_t( Workspace&& w = {} ) : work(w) { }

   inline constexpr
   workspace_opts_t( const Workspace& w ) : work(w) { }

   template< class matrix_t >
   inline constexpr
   workspace_opts_t( const matrix_t& A )
   : work( legacy_matrix(A).in_bytes() ) { }
};
```

## Workspace queries and workspace allocation:

Since workspace is optional in \<T\>LAPACK, we provide queries for obtaining the workspace sizes for external allocation. The query for a routine `foo( x, y, z, opts )` is the routines `foo_worksize( x, y, z, workinfo, opts )`, where the entry `workinfo` is output-only and contains all information necessary to allocate workspace before calling `foo( x, y, z, opts )`.

In what concern the use of the queries:

- Query before allocating the workspace outside \<T\>LAPACK, as opposed to assuming the workspace size. Or, let \<T\>LAPACK allocate the workspace for you. This helps user code to be always up-to-date with \<T\>LAPACK implementations. 

- Every routine `foo` that does internal allocation calls `foo_worksize`. This helps test the workspace allocation in place and avoids repetition of logic.

In what concerns allocation:

- Routines in \<T\>LAPACK should allocate space all at once, instead of letting multiple internal calls do that. One example is the QR factorization using elementary Householder reflectors. The level-2 routine (`geqr2`) needs extra space to apply the Householder reflectors, on every loop iteration, using `larf`. The allocation is done at `geqr2` for the larger space needed.

- `vectorOfBytes = std::vector<byte,byteAlloc>` is the type \<T\>LAPACK uses to store dynamically allocated memory inside each routine. Since we currently do not use any of the functionality from `std::byte` from C++17, we define `byte = unsigned char`. `byteAlloc` is a specific kind of allocator that, if needed, can be modified to attend requisites of certain systems.

- `alloc_workspace` is a utility function to help workspace allocation inside and outside \<T\>LAPACK routines. `alloc_workspace( vectorOfBytes& v, std::size_t lwork )` is used whenever the dynamic memory should be allocated. `alloc_workspace( vectorOfBytes& v, const workinfo_t& workinfo, const Workspace& opts_w )` allocates memory if `workinfo` is empty and, if not, tries to use `opts_w` as workspace.

- `deduce_work` is a simple helper to be used in routines that accepts workspace type to be informed through the interface instead of automatically deduced. See `larfb` and `unmqr`.

Assumptions:

1. Given a routine that requires workspace. If we increase the size of one of the inputs of a routine, the workspace needed for the new call is either the same or larger than the one we had at the beginning.

2. The workspace sizes do not depend on any entry of input arrays. Also, they do not depend on the value of scalar arguments.

3. The workspace sizes must depend on the size of arrays, and/or other sizes passed as optional arguments. If workspace sizes do not depend on the sizes given on input, one should use static memory internally in such a routine instead.

Design choices:

- The interface of the query mimics the interface of the original function. The difference is that the query returns `void` and has an output-only entry `workinfo_t& workinfo`. The structure `workinfo_t` holds information on workspace size, for contiguous workspaces, and the number of rows and columns, for non-contiguous workspaces. It also dictates the rule to compare two workspace sizes and find the minimal workspace configuration that contains both.

- Simple queries should be `inline constexpr` to allow for better optimization by the compiler. This could be especially useful for recursive routines, where there will be multiple calls for those routines.

- We should never annotate the result of a workspace query in the documentation. This is to force the users to use the query if allocation will be treated outside the routine. If a routine changes its implementation, the workspace may change, and then users' codes do not need to be changed.

- Workspace queries should call other workspace queries depending on the internal calls of each algorithm.

Notice that a workspace query may use constant references instead of non-const references in the interface. Example: `matrix_t& A` may be transformed into `const matrix_t& A`. The objective would be to prohibit the query to modify inputs by mistake. The downside here is that this idea creates slightly different interfaces between a routine and its workspace query.

## New traits in Matrix and Vector types:

There have been included multiple traits in the plugins for matrices and vectors.

1. `transpose_type< matrix_t >` is used to get the transposed type corresponding to `matrix_t`. This is useful for creating matrices of size m-by-n from a workspace of size n-by-m. See the example in `gehrd`.

2. `CreateImpl< matrix_t, int >` is the functor used by `Create< matrix_t >`. It must implement different operators `operator()` as described in the section [Matrix and vector creation](#matrix-and-vector-creation).

3. `matrix_type< array_t... >` is the matrix type that better matches the list of types `array_t...`. `vector_type< array_t... >` is the vector type that better matches the list of types `array_t...`. Notice that is possible to use matrix types in the list for `vector_type`, as well as vector types in the list for `matrix_type`. Those structures are the key to the automatic deduction of the type of workspaces. The `TLAPACK_PREFERRED_MATRIX` macro is used to define the default matrix and vector types in \<T\>LAPACK.

   ```C++
   template< typename matrixA_t, typename matrixB_t >
   struct matrix_type_traits< matrixA_t, matrixB_t >

   template< typename vecA_t, typename vecB_t >
   struct vector_type_traits< vecA_t, vecB_t >
   ```

   At least one of the plugin files must include those implementations.

## Additional changes:

- `legacyMatrix` is now part of \<T\>LAPACK abstract interface. This needed to be done because we now allow the automatic allocation of data types. Because of compatibility with BLAS++, `legacyMatrix` now has a template for size type.

- Examples and tests were updated accordingly. Since it is now possible to create matrices and vectors for a given data structure, tests are more generic and may be modified to accept Eigen and mdspan types.

- `one_t` was moved to `types.hpp`, and it is now called `StrongOne`.

- All workspace allocation was suppressed from the `legacy_api` wrappers.

PR #129, opened 2022-08-30
lu decomposition

alilotfi90:
add lu decomposition and tests

### DIFF
--- a/include/tlapack/lapack/getrf.hpp
+++ b/include/tlapack/lapack/getrf.hpp
@@ -1,5 +1,5 @@
 /// @file geqr2.hpp
-/// @author Weslley S Pereira, University of Colorado Denver, USA
+/// @author Ali Lotfi, University of Colorado Denver, USA
 /// Adapted from @see https://github.com/langou/latl/blob/master/include/geqr2.h
 //
 // Copyright (c) 2013-2022, University of Colorado Denver. All rights reserved.
@@ -55,6 +55,8 @@ template< class matrix_t >
 int getrf( matrix_t& A )
 {
     using idx_t = size_type< matrix_t >;
+    using T = type_t<matrix_t>;
+    using real_t = real_type<T>;
     // using pair  = pair<idx_t,idx_t>;
 
     // constants
@@ -72,7 +74,7 @@ int getrf( matrix_t& A )
     
     for(idx_t j=0;j<end;j++){
         for(idx_t i=j+1;i<m;i++){
-            assert( A(j,j) != 0 );
+            assert( A(j,j) != real_t(0) );
             A(i,j)=A(i,j)/A(j,j);
         }
         for(idx_t row=j+1;row<m;row++){
--- a/include/tlapack/lapack/getrf.hpp
+++ b/include/tlapack/lapack/getrf.hpp
@@ -41,8 +41,8 @@ namespace tlapack {
  *
  * @ingroup group_solve
  */
-template< class matrix_t >
-int getrf( matrix_t& A, std::vector<idx_t> &Piv)
+template< class matrix_t , class vector_t >
+int getrf( matrix_t& A, vector_t &Piv)
 {
     using idx_t = size_type< matrix_t >;
     using T = type_t<matrix_t>;
@@ -55,7 +55,7 @@ int getrf( matrix_t& A, std::vector<idx_t> &Piv)
 
     // check arguments
     tlapack_check_false( access_denied( dense, write_policy(A) ) );
-    tlapack_check( (idx_t) Piv.size() >= end);
+    tlapack_check( (idx_t) size(Piv)>= end);
     // quick return
     idx_t toswap = idx_t(0);
     if (m<=0 || n <= 0) return 0;
--- a/include/tlapack/lapack/getrf.hpp
+++ b/include/tlapack/lapack/getrf.hpp
@@ -55,7 +55,7 @@ int getrf( matrix_t& A, std::vector<idx_t> &Piv)
 
     // check arguments
     tlapack_check_false( access_denied( dense, write_policy(A) ) );
-
+    tlapack_check( (idx_t) Piv.size() >= end);
     // quick return
     idx_t toswap = idx_t(0);
     if (m<=0 || n <= 0) return 0;
--- a/include/tlapack/lapack/getrf.hpp
+++ b/include/tlapack/lapack/getrf.hpp
@@ -1,6 +1,5 @@
 /// @file getrf.hpp
 /// @author Ali Lotfi, University of Colorado Denver, USA
-/// Adapted from @see https://github.com/langou/latl/blob/master/include/geqr2.h
 //
 // Copyright (c) 2013-2022, University of Colorado Denver. All rights reserved.
 //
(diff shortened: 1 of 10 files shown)


include/tlapack/lapack/getrf.hpp
```diff
@@ -0,0 +1,92 @@
+/// @file geqr2.hpp
```
langou, line 1: change

include/tlapack/lapack/getrf.hpp
```diff
@@ -0,0 +1,92 @@
...
+ * @param[out] tau Real vector of length min(m,n).
+ *      The scalar factors of the elementary reflectors.
+ *      If
+ *          n-1 < m and
+ *          type_t<matrix_t> == type_t<vector_t>
+ *      then one may use tau[1:n] as the workspace.
+ * 
+ * @param work Vector of size n-1.
+ * 
+ * @ingroup geqrf
```
langou, line 52: change. + speak to Weslley, I do not think we should d ``group`` anyway.

include/tlapack/lapack/getrf.hpp
langou: Please do not comment this line, change it to 
@file getrf.hpp

include/tlapack/lapack/getrf.hpp
langou: remove this ``@ingroup geqrf`` line

include/tlapack/lapack/getrf.hpp
```diff
@@ -0,0 +1,115 @@
...
+        tlapack::swap(vect1,vect2);
+        // for (idx_t i = 0; i < m; ++i)
+        // {
+        //     std::cout << std::endl;
+        //     for (idx_t k = 0; k < n; ++k)
+        //         std::cout << std::setw(16) << A(i, k) << " ";
+        // }
+        // return pivot is zero
+        if (A(j,j)==real_t(0)){
+            return j+1;
```
langou, line 90: Speak to @weslleyspereira 
I thought we were returning index 0 base. 
So that would be return j;

include/tlapack/lapack/getrf.hpp
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+
+        
+        
+        
+        for(idx_t i=j+1;i<m;i++){
+            A(i,j)=A(i,j)/A(j,j);
```
langou, line 98: maybe use /= here
A(i,j) /= A(j,j);

include/tlapack/lapack/getrf.hpp
```diff
@@ -0,0 +1,115 @@
...
+        }
+        for(idx_t row=j+1;row<m;row++){
+            for(idx_t col=j+1;col<n;col++){
+                A(row,col)=A(row,col)-A(row,j)*A(j,col);
```
langou, line 102: maybe use -= here
A(row,col)-=A(row,j)*A(j,col);

langou:
The code is more like GETF0 than GETRF. You need to start using some BLAS and write some variants, but this is a good start

include/tlapack/lapack/getrf.hpp
weslleyspereira: Hi Julien. In this case it should be j+1. If we return j, we will have problems to identify zero pivots for the first iteration because 0 indicates a successful exit

include/tlapack/lapack/getrf.hpp
```diff
@@ -0,0 +1,91 @@
+/// @file getrf.hpp
+/// @author Ali Lotfi, University of Colorado Denver, USA
+/// Adapted from @see https://github.com/langou/latl/blob/master/include/geqr2.h
```
weslleyspereira, line 3: Did you adapt from this link?

include/tlapack/lapack/getrf.hpp
```diff
@@ -0,0 +1,91 @@
...
+ *      4- On and above the diagonal of A will be copied to U
+ *      
+ * @param[in,out] Piv is a k-by-1 vector where k=min(m,n)
+ * and Piv[i]=j where i<=j<=k-1, which means in the i-th iteration of the algorithm,
+ * the j-th row needs to be swapped with i 
+ *
+ * @ingroup getrf
+ */
+template< class matrix_t >
+int getrf( matrix_t& A, std::vector<idx_t> &P)
```
weslleyspereira, line 41: I suggest you modify the name of the variable to Piv

include/tlapack/lapack/getrf.hpp
```diff
@@ -0,0 +1,91 @@
...
+    using real_t = real_type<T>;
+
+    // constants
+    const idx_t m = nrows(A);
+    const idx_t n = ncols(A);
+    const idx_t end = std::min<idx_t>( m, n );
+
+    // check arguments
+    tlapack_check_false( access_denied( dense, write_policy(A) ) );
+
```
weslleyspereira, line 54: Please, check the size of Piv `tlapack_check( (idx_t) P.size() >= k );`

include/tlapack/lapack/getrf.hpp
```diff
@@ -0,0 +1,91 @@
...
+    for(idx_t j=0;j<end;j++){
+        // find pivot and swap the row with pivot row
+        toswap=j+iamax(tlapack::slice(A,tlapack::range<idx_t>(j,m),j));
+        P[j]=toswap;
+        auto vect1=tlapack::row(A,j);
+        auto vect2=tlapack::row(A,toswap);
+        tlapack::swap(vect1,vect2);
+        
+        // if nonzero pivot does not exist, return 
+        if (A(j,j)==real_t(0)){
```
weslleyspereira, line 68: Based on the LAPACK implementation, I would say that you only swap if the pivot isn't zero. So, could you please do the check before swapping?

include/tlapack/lapack/getrf.hpp
weslleyspereira: Hi Julien. Currently, we are using `@ingroup` to organize the routines in the documentation. Look the way they are split now: https://tlapack.github.io/. This is the same in LAPACK++: https://icl.bitbucket.io/lapackpp/.

(Yes, I put the documentation online! It was trivial to do.)

langou (commented):
Can you add getrf2.hpp? I do not think you did ``git add``

alilotfi90:
Sorry, just pushed it.

langou (commented):
This is missing a getrf(A0,Piv0); at the start where A0 is a slice of A (0:m, 0:n0).
I'll reexplain.

include/tlapack/lapack/getri.hpp
```diff
@@ -0,0 +1,249 @@
...
+            // first step of the algorithm, work1 holds x12
+            // work1 = -X22 * l21
+            std::vector<T> work1( n-j-idx_t(1) , T(0) );
+            tlapack::gemv(Op::NoTrans,T(-1), X22, l21,T(0), work1);
+            // std::cout<<nrows(X22)<<std::endl;
+            
+            //second line of the algorithm, work2 holds x21
+            // work2 = -u12 X22 / A(j,j)
+            std::vector<T> work2( n-j-idx_t(1) , T(0) );
+            tlapack::gemv(Op::ConjTrans,T(-1)/A(j,j), X22, u12,T(0), work2);
```
weslleyspereira, line 83: I think you should use `Op::Trans` instead of `Op::ConjTrans`. This may be the origin of the problem you had earlier.

include/tlapack/lapack/getri.hpp
```diff
@@ -0,0 +1,249 @@
...
+ *      4. On and above the diagonal of A will be copied to U
+ *      
+ * @param[in,out] Piv is a k-by-1 vector where k=min(m,n)
+ * and Piv[i]=j where i<=j<=k-1, which means in the i-th iteration of the algorithm,
+ * the j-th row needs to be swapped with i
+ *
+ * @ingroup group_solve
+ */
+template< class matrix_t, class vector_t >
+int getri( matrix_t& A, vector_t &Piv){
```
weslleyspereira, line 46: Piv is not an input. You invert matrix A, and the result is the matrix A inverted. Piv is a local variable of `getri`.

include/tlapack/lapack/getri.hpp
```diff
@@ -0,0 +1,249 @@
...
+            A(j,j) = T(1) / A(j,j);
+        }
+        else{
+            auto X22 = tlapack::slice(A,tlapack::range<idx_t>(j+1,n),tlapack::range<idx_t>(j+1,n));
+            auto l21 = tlapack::slice(A,tlapack::range<idx_t>(j+1,n),j);
+            auto u12 = tlapack::slice(A,j,tlapack::range<idx_t>(j+1,n));
+
+            // first step of the algorithm, work1 holds x12
+            // work1 = -X22 * l21
+            std::vector<T> work1( n-j-idx_t(1) , T(0) );
```
weslleyspereira, line 76: - You need 1 workspace of size n-1. I suggest you declare it before the loop for now. We will move it to the interface at some point. So, for every iteration, do `w = tlapack::slice( work, tlapack::range<idx_t>(0,n-j-1) );` and then work with `w`.

- I think the correct order of operations to optimize workspace should be:
  1. `w := - X22^T * u12 / A(j,j)`
  2. Save w on the line j of A.
  3. `w := -X22 * l21`
  4. Update `A(j,j)`
  5. Save w on the column j of A.

include/tlapack/lapack/getri.hpp
```diff
@@ -0,0 +1,249 @@
...
+        }
+    }
+    // for(idx_t j=n-idx_t(1);j!=idx_t(-1);j--){
+    //             if(Piv[j]>j){
+    //                 auto vect1=tlapack::col(A,j);
+    //                 auto vect2=tlapack::col(A,Piv[j]);
+    //                 tlapack::swap(vect1,vect2);
+    //             }
+    //     }
+    for(idx_t j=idx_t(0);j<n;j++){
```
weslleyspereira, line 102: This should be backwards as we discussed

include/tlapack/lapack/getri.hpp
weslleyspereira: Actually, @langou corrected me. getri indeed has an input-only Piv. Please, look on LAPACK.

```
 DGETRI computes the inverse of a matrix using the LU factorization
 computed by DGETRF.
```
 
So, the problem is that getri shouldn't do everything, but just use the precomputed L and U factors to obtain the inverse of A.

PR #135, opened 2022-09-29
Add LU decomposition level 0

alilotfi90:
add lu decomposition and tests

related to #129


test/src/test_getrf.cpp
```diff
@@ -0,0 +1,97 @@
...
+//
+// This file is part of <T>LAPACK.
+// <T>LAPACK is free software: you can redistribute it and/or modify it under
+// the terms of the BSD 3-Clause license. See the accompanying LICENSE file.
+
+#include <catch2/catch_template_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
+#include <tlapack/plugins/stdvector.hpp>
+#include <tlapack/plugins/legacyArray.hpp>
+#include <tlapack/lapack/getrf.hpp>
```
weslleyspereira, line 14: You need to add
`#include <tlapack/blas/swap.hpp>`
here

weslleyspereira, line 14: Some tests are failing because we do not have the wrappers to BLAS in `swap.hpp`. The wrappers are on `wrappers.hpp`. Maybe each wrapper should be in one file.

For now, I suggest you replace `#include <tlapack/lapack/getrf.hpp>` by `#include <tlapack/tlapack.hpp>` as you originally wrote.

weslleyspereira: I implemented the fix

alilotfi90:
Thank you Weslley.

Best,
Ali

PR #136, opened 2022-10-06
Adds LU recursive + LU inverse

weslleyspereira:
This PR adds a couple of things
    
- LU decomposition using a recursive algorithm (getrf2)
- Inversion of a matrix A solving the equation UXL=I, where A=LU (getri)
- It also adds an algorithm to perform the UL multiplication
- Adds tests for every of those algorithms


weslleyspereira:
TO-DO in my viewpoint:
- Change getrf2 to getrf_recursive and current getrf to getrf_level0, or something similar
- getri_methodC -> getri_methodUXLI, or something similar
- Create a getrf( ..., opts ), where opts have opts.method. This will help to keep a single interface to multiple variants. We should have opnly one test file for testing all variants. 
- work vector in getri should be at the interface of the routine. This is something that is changing with #128.

weslleyspereira:
I think this PR is ready for review and maybe merge.
- The 28 commits will be squashed to a single one. To improve readability of the history 
- The failing tests are related to updates on BLAS++ and LAPACK++. Unrelated to this current branch

PR #143, opened 2022-11-03
Improvements in the testing framework

weslleyspereira:
Improvements:

- Install Catch2 separately from \<T\>LAPACK in the CI:
  Catch2 is taking a big part __(on Windows and MacOS)__ of the time of the installation of \<T\>LAPACK in the CI. It is good to know the actual time our code is taking to install. Moreover, if one is developing code for \<T\>LAPACK, it will certainly install Catch2 and other packages only once. One of the CI cases will still test the automatic instalation of Catch2.

- Move the wrappers to each corresponding file. This allows users (and tests and examples in \<T\>LAPACK) to include only the files in \<T\>LAPACK that are part of each program even when linking with optimized BLAS and LAPACK libraries. Also, it reduce the compilation time for developers: if you change QR decomposition, only tests that use QR decomposition need to be rebuilt.


weslleyspereira:
This is related to https://github.com/tlapack/tlapack/discussions/144

PR #147, opened 2022-11-07
Improve tests and workspace queries

weslleyspereira:
- Use TEMPLATE_TEST_CASE instead of TEMPLATE_LIST_TEST_CASE. The advantage here is that the test messages now will show the type tested instead of an index respective to the type tested.

- Replace DYNAMIC_SECTION by INFO. Advantage: Avoid creating unnecessary sections.

- Use workspace queries on tests.

- Fix bugs on the workspace queries.

- In workspace queries, workinfo is [in,out]. This reduces the lines of code when performing multiple queries. See test_unmhr.hpp and test_gebd2.hpp.

### DIFF
--- a/include/tlapack/base/utils.hpp
+++ b/include/tlapack/base/utils.hpp
@@ -867,25 +867,31 @@ struct workinfo_t
     size_t m = 0; ///< Number of rows needed in the Workspace
     size_t n = 0; ///< Number of columns needed in the Workspace
 
+    /// Constructor using sizes
+    inline constexpr workinfo_t( size_t m = 0, size_t n = 0 ) noexcept
+    : m(m), n(n) {}
+
     /// Size needed in the Workspace
-    inline constexpr
-    size_t size() const { return m*n; }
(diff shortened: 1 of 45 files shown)


weslleyspereira:
> @weslleyspereira I'm not sure what the problem is with DYNAMIC_SECTION. I thought it was really nice that it wasn't just a huge test for an entire routine, but sectioned off into different smaller tests... then if one fails but the rest pass, it is much easier to isolate (this was bugged, but should be fixed in the later versions of catch2)

I think they were making my compilation times slower, but I don't remember well.

I see your point can come back with the `DYNAMIC_SECTION`, and now it is clear to me. Moreover, the Catch2 community says that `DYNAMIC_SECTION` "(...) can be useful with e.g. generators, or when creating a SECTION dynamically, within a loop".

I can propose a PR to come back with them. Also, I will change the recommendations at https://github.com/tlapack/tlapack/blob/master/CONTRIBUTING.md#directions-for-writing-tests

PR #148, opened 2022-11-17
Adds Eigen matrices on the testers

### LINKED ISSUE #88: We allow different matrix types, but always test with the same
for example, gemm can accept three different types for its three matrices, but the way we are testing now, we only test with the same matrix type (i think).

a similar issue is that we only test for arithmetic types of the same type of the matrix.

However, testing everything would quickly lead to a combinatorial explosion. Not sure what we should do.

weslleyspereira:
Closes #88.

- Adds TLAPACK_TEST_EIGEN to enable tests with Eigen matrices
- Implement missing functionality in the plugin for Eigen
- Add tests for plugin Eigen

After integration with Eigen is clean, we should add mdspan too.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -123,6 +123,7 @@ jobs:
         cmake -B build -G Ninja
         -D BUILD_TESTING=ON
         -D BUILD_BLASPP_TESTS=ON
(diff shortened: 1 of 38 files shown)


weslleyspereira:
cd58ebb adds tests for Kokkos mdspan

PR #150, opened 2022-12-01
Adds an example with Eigen3, testing multishift QR

weslleyspereira:
This is another example testing the multishift QR. This time, we compare with the implementations in Eigen.

- When the flag `EIGEN_USE_MKL_ALL` is activated, Eigen uses `gees` from MKL.
- I have added a Jupyter notebook to this example, so the results are easier to reproduce.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -22,6 +22,7 @@ env:
   blaspp_DIR: "${{github.workspace}}/blaspp"
   lapackpp_DIR: "${{github.workspace}}/lapackpp"
   mdspan_DIR: "${{github.workspace}}/mdspan"
+  eigen_DIR: "${{github.workspace}}/eigen"
   Catch2_DIR: "${{github.workspace}}/Catch2"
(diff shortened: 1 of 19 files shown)


weslleyspereira:
The test failing is due to a conflict on Eigen's function `size`. This was addressed at https://gitlab.com/libeigen/eigen/-/commit/9210e71fb378a0f1542272506dc2759b6c147237

weslleyspereira:
> The test failing is due to a conflict on Eigen's function `size`. This was addressed at https://gitlab.com/libeigen/eigen/-/commit/9210e71fb378a0f1542272506dc2759b6c147237

Fixed by ae31f84. There is still one issue related to the compilation of BLIS. This is unrelated to the current PR

weslleyspereira:
Here follows a few slides with the results using [cdffe56](https://github.com/tlapack/tlapack/pull/150/commits/cdffe56476067b349f314601663990ca1498ae03) in my machine. The configuration is described in the slides:

[performanceExamplesTLAPACK.pdf](https://github.com/tlapack/tlapack/files/10297115/performanceExamplesTLAPACK.pdf)

PR #152, opened 2022-12-10
Adds direction to larf

### LINKED ISSUE #151: larf currently considers v[0]=1
The current interface of larf is

```c++
void larf_worksize(
    side_t side, storage_t storeMode,
    vector_t const& v, const tau_t& tau, matrix_t& C,
    workinfo_t& workinfo, const workspace_opts_t<>& opts = {} );
```

It assumes `v[0]=1`. The LAPACK version uses the whole vector v. The change was done so that v[0] doesn't need to be changed to 1 whenever one wants to apply the reflector. There is one problem that was not predicted: in the QL and RQ decompositions, the 1 is at the very end of the vector v.

I propose a solution that is the one used in larfb: add a parameter "direction". If direction is forward then `v = [ 1 x ]`, if direction is backward then `v = [ x 1 ]`.

I will propose the PR.

weslleyspereira:
Closes #151

### DIFF
--- a/include/tlapack/lapack/agressive_early_deflation.hpp
+++ b/include/tlapack/lapack/agressive_early_deflation.hpp
@@ -433,13 +433,13 @@ namespace tlapack
                 auto work2 = workspace_opts_t<>(Wv_aux);
                 
                 auto TW_slice = slice(TW, pair{0, ns}, pair{0, jw});
-                larf(Side::Left, v, conj(tau), TW_slice, work2);
+                larf(Side::Left, forward, v, conj(tau), TW_slice, work2);
                 
                 auto TW_slice2 = slice(TW, pair{0, jw}, pair{0, ns});
-                larf(Side::Right, v, tau, TW_slice2, work2);
+                larf(Side::Right, forward, v, tau, TW_slice2, work2);
                 
                 auto V_slice = slice(V, pair{0, jw}, pair{0, ns});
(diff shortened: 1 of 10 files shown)


langou:
This is not urgent but I think that

(1) A similar `direction` flag can be used in `larfg`, this would enable, in `geql2` to not have to use the `alpha` and `vector` representation in `larfg` but simply gives the full vector as is done in `geqr2`.

(2) in addition to implement the `direction` flag in `larfg`+`larf`, we will also want to implement the `storev` flag in `larfg`+`larf`. This way we can remove the conjugation of v in the  `gelq2` code at lines
https://github.com/tlapack/tlapack/blob/78dc01b426af3ca57be127c0a76b1e160fcb1ca6/include/tlapack/lapack/gelq2.hpp#L116
and lines
https://github.com/tlapack/tlapack/blob/78dc01b426af3ca57be127c0a76b1e160fcb1ca6/include/tlapack/lapack/gelq2.hpp#L127
And in a similar way, we will be able to remove the conjugation flag in `gerq2` code (which we do not have yet).

(3) That way, we would have specialized `larfg`+`larf` for the cases
QR is `direction`=`Forward`, `storev` = `Columnwise`
QL is `direction`=`Backward`, `storev` = `Columnwise`
LQ is `direction`=`Forward`, `storev` = `Rowwise`
RQ is `direction`=`Backward`, `storev` = `Rowwise`
and I think the four codes would be more readable.

PR #157, opened 2023-01-04
Try to make \<T\>LAPACK work with GMP multiprecision type

weslleyspereira:
- [x] Replace auto keyword by explicit type


weslleyspereira:
How is the `auto` keyword being used right now in \<T\>LAPACK:

1. To get new matrices and vectors returned by
    a. by `legacy_matrix` and `legacy_vector` routines in the wrappers to optimized BLAS.
    b. slicing a matrix or vector.
    c. the functor `Create<array_t>(...)`.
2. In the return type of `asum, dot, dotu, nrm2, lange, lanhe, lansy, lantr`.
3. In the Legacy API, to avoid verbosity when creating the wrappers.

Item (1) is essential to build generic algorithms that work with different matrix and vector classes.

Item (2) enables overloading in mixed precision. E.g., one may want to receive a vector in half precision and compute its norm in single precision to have better accuracy. \<T\>LAPACK allows overloading to encompass such scenario.

Item (3) is good for practical reasons. It reduces the verbosity in the wrappers of Legacy API.

-----

Why are we avoiding the usage of `auto`?
1. When the type is know and not too verbose, it is better to write it down. It simplifies the understanding of the code, and reduces the work that the compiler needs to do.
2. To avoid activating lazy evaluation when we don't want. Some types may use lazy evaluation to compute:
    ```c++
    auto x = y + z;
    T w = x * x;
    ```
    so that `y + z` will actually be computed twice. One example is the `mpf_class` from the GNU multiprecision library.

PR #176, opened 2023-01-30
Adds storageMode on larfg and larf

### LINKED ISSUE #84: larfb does not check bounds
Maybe @ntebeje could help with that.

weslleyspereira:
This PR adds more functionality to `larfg` and `larf` as follows:

- `larfg` now is able to store the Householder vectors in two ways: rowwise and columnwise. The latter is the functionality that is in LAPACK, used in QR, LQ and a large number of other routines. The rowwise mode is useful for RQ, QL, bidiagonalization, tridiagonalization, etc. In the rowwise mode, the conjugate of `v` is stored instead of `v` itself. See details in the documentation of `larfg`.

- `larf` has two new parameters: storeMode and direction. The first compatibilizes with the one from `larfg`. The second allows for the use of Householder vectors v = [1 x] and v = [x 1] more easily. That makes the algorithms for RQ and QL to be very similar to QR and LQ.

- `larf` has a new interface that receives `x, C0, C1` instead of `direction, v, C`. The new interface is the one that holds the logic behind the application of the Householder reflection, and the latter is basically a wrapper. The new interface generalizes the types of data one can apply a reflection. See the documentation for the details.

Additionally, we make the following changes:

- Adds a tester for LARF and LARFG.

- This PR closes #84, which is related to LARFB.

- When slicing a vector v or matrix A, allow the use of ranges (size(v),size(v)), (nrows(A),nrows(A)) and (ncols(A),ncols(A)). Notice that we still do not allow ranges like (size(v),size(v)+1), (nrows(A),nrows(A)+1). This change avoids ugly code like `(i < m) ? range(i,m) : range(0,0)` that was used just for obtaining a valid empty vector or matrix. It is worthy noticing that the matrix and vector types we support accept the new allowed ranges. 

- Adds the target `check-format-of-hpp-cpp-files` to the CMake build system. This checks if all files are correctly formatted accordingly to the rules in `.clang-format`. Mind that the check depends on the ClangFormat version installed in the system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,6 +215,15 @@ if(GIT_FOUND AND EXISTS "${PROJECT_SOURCE_DIR}/.git" AND CLANG_FORMAT_FOUND)
       2> format-all-hpp-cpp-files.err
     WORKING_DIRECTORY ${TLAPACK_SOURCE_DIR}
   )
+  add_custom_target( check-format-of-hpp-cpp-files
+    COMMAND
+      ${GIT_EXECUTABLE} ls-tree -r HEAD --name-only |
+      grep -E '\(config/|examples/|include/|src/|test/include/|test/src/\).*\\.\(hpp|cpp|h.in|c\)$$' |
+      xargs ${CLANG_FORMAT_EXECUTABLE} -i -style=file --dry-run
+      > format-all-hpp-cpp-files.log
+      2> format-all-hpp-cpp-files.err
+    WORKING_DIRECTORY ${TLAPACK_SOURCE_DIR}
+  )
 endif()
 
 #-------------------------------------------------------------------------------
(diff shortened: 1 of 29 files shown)


weslleyspereira:
Updated the description. I think this PR is ready to review

langou (approved):
I find this PR great. And elegant. (1) This makes using `larfg` and `larf` a little bit more verbose, but I find that the added feature is worth the verbosity. Currently LAPACK does not have this feature and the LAPACK code for xGELQ is "ugly" and performs more passes on the vectors `v` than needed. (2) I do like the fact that now the interface of `larf` is more similar to the interface of `larfb` which makes sense, but is not something that you currently have in LAPACK. (3) I like the design where we have `alpha` (or `beta`) attached to `x` (or `y`) or not. We often have contiguous data, and sometimes we do not. So it is good to have both.

PR #178, opened 2023-01-31
Implements the reciprocal scaling

### LINKED ISSUE #139: Implement reciprocal scaling
Reciprocal scaling is `x := x/alpha`, where x is an array. This operation is, in most cases, preferable to more `x := (1/alpha)*x`. For instance, the former has less accumulation error.

The technical report on the BLAS Standard, https://netlib.org/blas/blast-forum/blas-report.pdf, proposes the reciprocal scaling to be part of the BLAS. The current LAPACK code has its implementation of RSCL, but it relies on applying the BLAS SCAL to (1/alpha). MKL, for instance, has its own implementation of RSCL.

I think we should have a tlapack::rscl. I can propose the PR. Also, I can propose the PR in LAPACK++ so we can access code from MKL and other optimized libraries.

weslleyspereira:
Closes #139.

### DIFF
--- a/include/tlapack.hpp
+++ b/include/tlapack.hpp
@@ -86,6 +86,7 @@
 #include "tlapack/lapack/lassq.hpp"
 #include "tlapack/lapack/lauum_recursive.hpp"
 #include "tlapack/lapack/lu_mult.hpp"
+#include "tlapack/lapack/rscl.hpp"
 #include "tlapack/lapack/transpose.hpp"
 
 // SVD
--- a/include/tlapack/lapack/getrf_level0.hpp
+++ b/include/tlapack/lapack/getrf_level0.hpp
@@ -89,6 +89,7 @@ int getrf_level0(matrix_t& A, vector_t& Piv)
 
         // divide below diagonal part of j-th column by the element on the
         // diagonal(A(j,j))
+        // This is a level 0 routine, so we do not use tlapack::rscl.
         for (idx_t i = j + 1; i < m; i++) {
             A(i, j) /= A(j, j);
         }
--- a/include/tlapack/lapack/getrf_recursive.hpp
+++ b/include/tlapack/lapack/getrf_recursive.hpp
@@ -13,9 +13,9 @@
 #include "tlapack/base/utils.hpp"
 #include "tlapack/blas/gemm.hpp"
 #include "tlapack/blas/iamax.hpp"
-#include "tlapack/blas/scal.hpp"
 #include "tlapack/blas/swap.hpp"
 #include "tlapack/blas/trsm.hpp"
+#include "tlapack/lapack/rscl.hpp"
 
 namespace tlapack {
 
@@ -110,7 +110,7 @@ int getrf_recursive(matrix_t& A, vector_t& Piv)
         // by the previous comment, we can safely scale all elements of 0th
         // column by 1/A(0,0)
         auto l = slice(A, range<idx_t>(1, m), 0);
-        scal(real_t(1) / A(0, 0), l);
+        rscl(A(0, 0), l);
 
         return 0;
     }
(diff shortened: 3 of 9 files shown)


langou (commented):
Nice!

langou:
Maybe propagate to:
https://github.com/tlapack/tlapack/blob/0377a585a2e15653e4e6006a4aae53297b7f3ba6/include/tlapack/lapack/getrf_level0.hpp#L92

langou:
Here then:
https://github.com/tlapack/tlapack/blob/0377a585a2e15653e4e6006a4aae53297b7f3ba6/include/tlapack/lapack/getrf_recursive.hpp#L102

weslleyspereira:
> Here then:
> 
> https://github.com/tlapack/tlapack/blob/0377a585a2e15653e4e6006a4aae53297b7f3ba6/include/tlapack/lapack/getrf_recursive.hpp#L102

This was done in 3d947adc339102ff8771497828904e3eac8b16e2

langou:
Excellent then!

weslleyspereira:
[90496b7](https://github.com/tlapack/tlapack/pull/178/commits/90496b79cad1c91678be4022dca0d470bcee7b38) fixes an issue created by the new implementation.

The current implementation of `rscl` has two behaviors:
1. If `alpha` is real, then `x[i] /= alpha`.
2. If `alpha` is complex, then call `scal(real_type<alpha_t>(1) / alpha, x)`.

weslleyspereira:
I am putting this as a draft since we are still missing the scaling of `1/alpha` when alpha is complex.

weslleyspereira:
That is the current design of the algorithm I propose:

If alpha is real, then this routine is equivalent to scal(1/alpha, x). This
is done without overflow or underflow as long as the final result x/a does
not overflow or underflow.

If alpha is complex, then we use the following algorithm:
1. If the real part of alpha is zero, then we scale by the reciprocal of
the imaginary part of alpha.
2. If not, see if either real or imaginary part is greater than safeMax. If
so, do proper scaling.
3. If not, we can compute the reciprocal of real and imaginary parts of
1/alpha without NaNs. If both components are in the safe range, then we can
do the reciprocal without overflow or underflow. Otherwise, we scale by the
safe range.

weslleyspereira:
In the following, I present two matrices and use LAPACK's getf2 to factorize them. We compare getf2 in LAPACK 3.11 with getf2 using the new RSCL we propose in this PR.

1. Take the matrix:
```fortran
A = [  b+i  1  ]
    [  b    b  ]
```

For this matrix, LU with pivoting does not permute the row for this matrix and its exact LU factorization is 
```fortran
L = [  1      0 ]       U = [  b+i   1       ]
      [  l21   1  ]              [  0      u22  ]
```
where 
```
 l21 = b / (b + i )
 u22 = b / (b+i) * ( b+i-1 )
```

2. Now, let us use floating-point arithmetic and set `b` as a large number for this aritmetic. For example, let us use single precision and `b = 2^75`.

In this case, LAPACK's getf2 using the inline seven-line reciprocal scaling algorithm returns the factorization:
```
L = [  1      0 ]       U = [  b+i   1   ]
      [  1       1 ]              [  0      b  ]
```
whereas the new reciprocal scaling algorithm
```
L = [  1          0 ]       U = [  b+i   1   ]
      [  l21       1 ]              [  0      b  ]
```
The relative error is very small: 1.528233e-23, so no problems here. No problems because the condition number is small.
__We obtain the correct `b/(b+i)` when using the new reciprocal scaling algorithm. Note that, in finite single precision, `b/(b+i)` rounds up to `1 - i/b`, not to `1`.__

2. Now, let me we try the matrix:
```
b+I b
  b b
```
The condition number here is awful, larger than 2b. In this case, LAPACK's getf2 returns the factorization:
```
b+I b
  1 0
```
We still have a small relative error. However, the LU factorization returns a singular matrix since there is a 0 in the diagonal of U. If we try to solve a linear system we may end up in a division by zero which raises an exception.

The correct LU factorization is 
```
    b+I b
b/(b+i) I
```
which is the one we obtain by using the algorithm I propose in this PR.

PR #189, opened 2023-02-20
Reduce a hermitian matrix to real symmetric tridiagonal form

weslleyspereira:
This PR is a continuation of @skylarstjohns' work. Thanks @skylarstjohns!

OBS: Still missing a test for the Upper triangular case.

### DIFF
--- a/include/tlapack/lapack/hetd2.hpp
+++ b/include/tlapack/lapack/hetd2.hpp
@@ -1,0 +1,188 @@
+/// @file hetd2.hpp
+/// @author Skylar Johns, University of Colorado Denver, USA
+//
+// Copyright (c) 2021-2023, University of Colorado Denver. All rights reserved.
+//
+// This file is part of <T>LAPACK.
+// <T>LAPACK is free software: you can redistribute it and/or modify it under
+// the terms of the BSD 3-Clause license. See the accompanying LICENSE file.
+
+#ifndef TLAPACK_HETD2_HH
+#define TLAPACK_HETD2_HH
+
+#include "tlapack/base/utils.hpp"
+#include "tlapack/blas/symv.hpp"
+#include "tlapack/lapack/larf.hpp"
+#include "tlapack/lapack/larfg.hpp"
+
+namespace tlapack {
+
+/** Worspace query of geqr2()
+ *
+ * @param[in] A m-by-n matrix.
+ *
+ * @param tau Not referenced.
+ *
+ * @param[in] opts Options.
+ *
+ * @param[in,out] workinfo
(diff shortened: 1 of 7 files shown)


include/tlapack/lapack/hetd2.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+    // quick return
+    if (n <= 0) return 0;
+
+    if (uplo == Uplo::Upper) {
+        //
+        // Reduce upper triangle of A
+        //
+
+        // Only access real part of the main diagonal
```
thijssteel, line 70: A should be Hermitian, so the entry should already be real. This line seems more like an extra check to avoid problems, the comment should be changed to reflect that.

include/tlapack/lapack/ung2l.hpp
```diff
@@ -0,0 +1,138 @@
+/// @file ung2l.hpp generates all or part of the unitary matrix Q from a QL
```
thijssteel, line 1: Ah damn, i had also implemented this without realizing you made it already. I like yours better, changing larf was a good idea.

test/src/test_hetd2.cpp
```diff
@@ -0,0 +1,150 @@
...
+        auto A = new_matrix(A_, n, n);
+        std::vector<T> Q_;
+        auto Q = new_matrix(Q_, n, n);
+        std::vector<real_t> E(n - 1), D(n);
+        std::vector<T> tau(n - 1);
+
+        // Fill A with random values
+        for (idx_t j = 0; j < n; ++j)
+            for (idx_t i = 0; i < n; ++i)
+                A(i, j) = rand_helper<T>();
```
thijssteel, line 67: perhaps you can fill the parts of A that should not be accessed with NaN or something and then later check that they were not changed.

thijssteel (approved):

include/tlapack/lapack/hetd2.hpp
langou: We are trying to respect LAPACK where the convention is something like: "yes the matrix is Hermitian, yes diagonal entries are real, if a diagonal entry is not real, this is meant to mean "only use the real part and assume imaginary is intended to be zero but might not be set as such" of this entry." 
https://github.com/Reference-LAPACK/lapack/blob/7ac31faec5b5b02a2591414f30b9aed3005b32a8/BLAS/SRC/zherk.f#L131-L133

test/src/test_hetd2.cpp
weslleyspereira: Thanks! That is a good idea.

thijssteel:
Completed by #346

PR #191, opened 2023-02-21
make TT optional in gelqf

### LINKED ISSUE #187: Suggested interface change for gelqf
Right now the routine gelqf has the signature:

gelqf( A, TT, opts)

TT stores the upper triangular factors and also serves as workspace. Allowing them to be reused later.

I don't like this for two reasons:
- It makes the interface gelqf incompatible with gelq2 and makes the legacy interface more difficult
- TT depends on nb, which is an optional argument. Essentially, to make good code, this means that nb is no longer optional.

Proposed change:
- Make TT an optional argument and replace the mandatory interface with tau

thijssteel:
closes #187 

In the current implementation, we have different interfaces for the LQ factorizations:
```c++
gelq2(A, tau, opts);
gelqf(A, TT, opts);
```
As outlined in the issue, TT depends on nb, which is an optional argument. Essentially, to make good code, this means that nb is no longer optional, because the user would need to check that TT actually matches nb.

This PR changes the interface of gelqf to match gelq2, i.e.
```c++
gelqf(A, tau, opts);
```

`opts` now contains a pointer to a matrix, if it is present, it is used an no new matrix is constructed from the workspace. If is it not present, a new matrix is constructed.

EDIT: After some discussion, the optional TT arguments was deemed to be too complicated. The routine is now split into gelqf and gelqt.


thijssteel:
The failure in the build is from mdspan. It seems like the matrix returned by the create implementation is not necessarily exactly matrix_t. I'm not sure how to work around that. Or maybe it's something else :shrug: templates are still a bit of a black box for me

include/tlapack/lapack/gelqf.hpp
```diff
@@ -128,29 +136,41 @@ int gelqf(matrix_t& A,
...
     }();
 
-    // Options to forward
-    auto&& gelq2Opts = workspace_opts_t<>{work};
+    // TT serves two functions, it stores the triangular factors
+    // in the compact WY representation and the parts that do not
+    // store triangular factors yet are used as workspace.
+    // If passed as an optional arguments, it is used so that the
+    // triangular factors can be reused later. Otherwise, workspace is used.
+    matrix_t TT = [&]() {
```
weslleyspereira, line 155: I think the problem is here. Try replacing `matrix_t` by `auto`. I explain:

- When `*opts.TT` is returned, the type of matrix `TT` is `matrix_t`. We're fine.
- When `new_matrix(work, m, nb, sparework)` is returned, we have whatever `new_matrix` returns. In case of Eigen, `new_matrix` returns a `Eigen::Map<matrix_t>`. This is the class in Eigen that maps pointers to raw data to a Eigen matrix.

It is listed in [CONTRIBUTING.md](https://github.com/tlapack/tlapack/blob/master/CONTRIBUTING.md) at "We recommend the usage of auto in the following cases:".

thijssteel: I had the same idea, but the problem is slightly earlier. The compiler needs to know the return type of the lambda function. In the situation outlined in contributing.md, the return type is determined at compile time.

weslleyspereira (approved):
Other than the fix about the `auto` keyword, I liked the changes. Thanks!

Note: We are planning to put a top-level interface `gelqf(A,tau,opts)` that will call those two variants based on `opts.variant`. This is a step towards having this top-level interface.

include/tlapack/lapack/gelqf.hpp
thijssteel: having a top level routine might solve it, then it can be encapsulated in two instantiations of the same function

thijssteel:
I ended up going with an if and no extra function, because if you don't want TT you can avoid calculating the triangular factors if the remaining part is small and switch to unblocked code.

include/tlapack/lapack/gelqf.hpp
```diff
@@ -26,14 +26,14 @@ struct gelqf_opts_t : public workspace_opts_t<> {
     inline constexpr gelqf_opts_t(const workspace_opts_t<>& opts = {})
         : workspace_opts_t<>(opts){};
 
-    idx_t nb = 32;  ///< Block size
```
weslleyspereira, line 29: Not a big deal here, but I want to mention that `///<` is the way Doxygen recognizes comments for attributes of classes.

weslleyspereira (commented):
It is better to put you, @thijssteel, as the author of gelqf.hpp. @yuxincai11 is the author of the code that is now in gelqt.hpp. Does it make sense? If not, leave like that.

weslleyspereira:
Looks good to me. Thanks!

PR #195, opened 2023-02-23
Add blocked qr implementation

thijssteel:
Adds blocked qr implementation in the same style as PR #191 . Also updates `unmqr` so that it can optionally reuse `TT`.


thijssteel:
An important thing to keep in mind for future implementations: writing something like `auto A = *opts.A;` leads to a copy if the matrix is an Eigen matrix. Not entirely sure why.

weslleyspereira:
> An important thing to keep in mind for future implementations: writing something like `auto A = *opts.A;` leads to a copy if the matrix is an Eigen matrix. Not entirely sure why.

Thanks for the warning! It makes sense, and I can explain. Since `*opts.A` is an `Eigen::Matrix<...>`, the compiler will replace `auto` by `Eigen::Matrix<...>` which is the best it can do. Since each object of the class Eigen::Matrix<...> owns its data pointer, there is no way `A` and `*opts.A` have access to the same pointer. That is why they implement the assignment operator as a copy. It is also explained here https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html. Look for  "Assignment and resizing".

The best we could do in this case is to use `auto& A = *opts.A;`. Now `A` is a reference and looks at the same position in memory as `*opts.A`.

I will add this to the FAQ in the wiki.

PR #196, opened 2023-03-14
Removes Op:Conj from gemv, trmv, trsm and larf

weslleyspereira:
This PR removes `Op::Conj` from gemv, trmv and trsm.

`Op::Conj` was used to indicate "transpose without conjugate". It may look like a good idea to have those functionalities in BLAS. However, this creates issues when linking with optimized BLAS, since the new flags aren't part of the standard.

More specifically, the issue is that the call `tlapack::gemv(Op::Conj, ... )` sometimes ends up in the wrapper for the optimized BLAS gemv which does not accept `Op::Conj`. We could as well make the wrapper call the C++ implementation of gemv if `trans == Op::Conj`, I think this could be too smart for a wrapper. The wrapper was designed to be just a call for the optimized implementation.

### DIFF
--- a/include/tlapack/blas/gemv.hpp
+++ b/include/tlapack/blas/gemv.hpp
@@ -31,8 +31,7 @@ namespace tlapack {
  *     The operation to be performed:
  *     - Op::NoTrans:   $y = \alpha A   x + \beta y$,
  *     - Op::Trans:     $y = \alpha A^T x + \beta y$,
- *     - Op::ConjTrans: $y = \alpha A^H x + \beta y$,
- *     - Op::Conj:  $y = \alpha conj(A) x + \beta y$.
+ *     - Op::ConjTrans: $y = \alpha A^H x + \beta y$.
  *
  * @param[in] alpha Scalar.
  * @param[in] A $op(A)$ is an m-by-n matrix.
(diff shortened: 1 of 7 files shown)


weslleyspereira:
Thanks for explaining better, @langou!

One additional remark: the wrappers for `gemv, trmv, trsv` right now are not prepared to deal with `Op::Conj`. Therefore, we may end up in an error when calling `gemv( Op::Conj, ... )` whenever we link we optimized BLAS libraries. Regardless, we should make a change in the code related to that.

thijssteel:
I would also like to add to this that in my experience, you don't get that big of a performance gain from anything less than lvl 3 to the point that I consider them syntactic sugar. I made this plot for my thesis to illustrate that.

image

I we're not calling optimized BLAS for lvl 1 or 2, it shouldn't significantly impact performance.

weslleyspereira:
> I would also like to add to this that in my experience, you don't get that big of a performance gain from anything less than lvl 3 to the point that I consider them syntactic sugar. I made this plot for my thesis to illustrate that.
> 
> image
> 
> I we're not calling optimized BLAS for lvl 1 or 2, it shouldn't significantly impact performance.

Thanks!

Do you know if this is still valid in double precision? 

What about on accelerators? For instance, if we use CuBLAS on a NVIDIA GPU. We are not prepared for accelerators, I am just wondering if you have any tests. Thanks again!

langou:
Yes, about level 1 and level 2, I thought that a compiler would be able to give good performance from a one loop or two loops, but actually I found that

weslleyspereira:
#226 had a better solution, i.e., keep flag Op::Conj and solves the problem with the wrapper for BLAS++ indicated at https://github.com/tlapack/tlapack/pull/196#issue-1624419030.

PR #207, opened 2023-03-30
Introduces iamax that uses lambda function for abs

weslleyspereira:
`iamax` had already options for NaN checking. This is an upgrade in the functionality of `iamax`. Now, we can pass the absolute value to be used `iamax`. For instance, we can call `iamax` using the usual complex norm

```c++
// x is a std::vector< std::complex<double> >
iamax_opts_t opts([](const std::complex<double>& x) { return std::abs(x); });
size_t idx = iamax( x, opts );
```

The LU recursive code now uses the usual complex norm. This is just so we can test the new functionality. We may need to add more tests. Moreover, we should decide how to use (or not use) this new functionality.

### DIFF
--- a/include/tlapack/blas/iamax.hpp
+++ b/include/tlapack/blas/iamax.hpp
@@ -17,7 +17,34 @@
 namespace tlapack {
 
 /**
- * @brief Return $\arg\max_{i=0}^{n-1} \left(|Re(x_i)| + |Im(x_i)|\right)$
+ * Options for iamax.
+ *
+ * Initialize using a lambda function (C++17 or higher):
+ * ```c++
+ * iamax_opts_t opts( [](const T& x) { return my_abs(x); } );
+ * ```
+ * or using a functor:
+ * ```c++
+ * struct abs_f {
+ *    inline constexpr real_type<T> operator()(const T& x) const {
+ *       return my_abs(x);
+ *   }
+ * };
+ * abs_f absf;
+ * iamax_opts_t<abs_f> opts( absf );
+ * ```
+ */
+template <class abs_f>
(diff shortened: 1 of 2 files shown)


langou:
Oh my. I like it. But oh my, do we go out there. 

So, for complex arithmetic, the pivoting strategy of \<T\>LAPACK GEPP is not the same as the one in LAPACK GEPP.

Fun! I like it.

That being said, I am glad there is a ``knob`` to change this back.

langou:
Yes, this made me think of coding an abstract "reduce" in <T>LAPACK where you would give the binary and associative operation to the "reduce". For example NORM2 is a reduction with LAPY2. LANGE('M') is a reduction with a MAX(). IAMAX can also be viewed as a reduction but it is a little more complicated to describe. Etc.

weslleyspereira:
> I thought we were already using C++17?

I tried to compile with `-std=c++14` here. In fact, oops, I wasn't being to careful and the following code actually only works on C++17:

https://github.com/tlapack/tlapack/blob/46b2ce634790f8cec284291fd42241d3f0f6eea2/include/tlapack/base/utils.hpp#L48-L58

If we remove those `inline`, we comprise with C++14. Eigen3 plugin currently requires C++17 because it was complicated to avoid the `if constexpr`.

This said, I am convinced we should move to C++20 for `if constexpr`, CTAD (Class Template Argument Deduction) and concepts. In my view, we should
1. use all of C++17 since it has large compiler support,
2. keep track of what we are using from C++20.

weslleyspereira:
I will merge this one. (Because I have just needed it in a test I am doing now!)
We can easily revert this one if needed.

PR #210, opened 2023-04-03
Bidiagonal reduction

thijssteel:
This PR has several things related to bidiagonal reductions (preprocessing for SVD)

- Finishes Yuxin's code so that it can also handle matrices with n < n
- Introduces `ungbr` to generate unitary factors of the bidiagonal reduction
- Introduces `gebrd`, a blocked version of `gebd2`


weslleyspereira:
> I can't reproduce the compilation errors on my machine. I'll try to come back to this later

May be related to implicit type cast. I saw the problem is related to Eigen::half and it doesn't do implicit cast.

include/tlapack/lapack/conjugate.hpp
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @ingroup auxiliary
+ */
+template <class vector_t>
+void conjugate(const vector_t& x)
+{
+    using idx_t = size_type<vector_t>;
+
+    for (idx_t i = 0; i < size(x); ++i) {
+        x[i] = conj(x[i]);
```
weslleyspereira, line 30: The issue is that `std::conj( x )` always returns a complex number regardless if x is real or complex. So, add `using tlapack::conj;` to this function. It is explained in `utils.hpp`.

thijssteel: The problem wasn't the std::conj, but rather that the vector was declared as const, so of course I couldn't modify it :facepalm: Those are a few hours of my life I'll never get back.

include/tlapack/lapack/ungbr.hpp
```diff
@@ -0,0 +1,224 @@
...
+#ifndef TLAPACK_UNGBR_HH
+#define TLAPACK_UNGBR_HH
+
+#include "tlapack/base/utils.hpp"
+#include "tlapack/lapack/unglq.hpp"
+#include "tlapack/lapack/ungqr.hpp"
+
+namespace tlapack {
+
+enum class QorP : char { Q = 'Q', P = 'P' };
```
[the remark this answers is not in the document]
thijssteel: I don't really mind the `QorP` enum. I know that it's relatively specific, but that's why I included it in the file and didn't put it into something general. That being said, the functions really don't share a lot of functionality beyond their interface so I guess separate functions are reasonable.

PR #213, opened 2023-04-04
Protect const entry access in legacyMatrix

### LINKED ISSUE #212: Const matrices/vectors are still modifyable
Currently the access operators in our matrices are defined as follows:

```c++
    inline constexpr T& operator()(idx_t i, idx_t j) const noexcept
    {
        return (layout == Layout::ColMajor) ? ptr[i + j * ldim]
                                            : ptr[i * ldim + j];
    }
```

Because the function is defined as `const`, we can do something like this

```c++
template< class matrix_t>
void foo( const matrix_t& A  ){
     A(0,0) = 0;
}
```
i.e. we can modify a const matrix. While valid c++ syntax, I found some sources that consider this bad practice.
In the scientific software course I'm TA for, the students are taught to do something like this:

```c++
    inline constexpr T& operator()(idx_t i, idx_t j) noexcept
    {
        return (layout == Layout::ColMajor) ? ptr[i + j * ldim]
                                            : ptr[i * ldim + j];
    }

    inline constexpr const T& operator()(idx_t i, idx_t j) const noexcept
    {
        return (layout == Layout::ColMajor) ? ptr[i + j * ldim]
                                            : ptr[i * ldim + j];
    }
```

If the matrix is const, this returns a const reference so that it cannot be modified.

The change is not nearly that simple though. We would also need to change the plugins, because a slice of a const matrix would need to return a const slice, ... I think there can be some discussion whether or not it's worth the effort.

weslleyspereira:
Closes #212.

### DIFF
--- a/include/tlapack/base/legacyArray.hpp
+++ b/include/tlapack/base/legacyArray.hpp
@@ -38,7 +38,17 @@ struct legacyMatrix {
 
     static constexpr Layout layout = L;
 
-    inline constexpr T& operator()(idx_t i, idx_t j) const noexcept
+    inline constexpr const T& operator()(idx_t i, idx_t j) const noexcept
+    {
+        assert(i >= 0);
+        assert(i < m);
+        assert(j >= 0);
+        assert(j < n);
+        return (layout == Layout::ColMajor) ? ptr[i + j * ldim]
+                                            : ptr[i * ldim + j];
+    }
+
+    inline constexpr T& operator()(idx_t i, idx_t j) noexcept
     {
         assert(i >= 0);
         assert(i < m);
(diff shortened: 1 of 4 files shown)


thijssteel:
I think this doesn't really solve the issue, but rather moves it. You can no longer do this
```c++
void foo(const matrix_t& A){
   A(0,0) = 0;
}
```
But you can still do (at least i think you can)
```c++
void foo(const matrix_t& A){
   auto B = slice( A, pair{0,1}, pair{0,1}); 
   B(0,0) = 0;
}
```

weslleyspereira:
> I think this doesn't really solve the issue, but rather moves it. You can no longer do this
> 
> ```c++
> void foo(const matrix_t& A){
>    A(0,0) = 0;
> }
> ```
> 
> But you can still do (at least i think you can)
> 
> ```c++
> void foo(const matrix_t& A){
>    auto B = slice( A, pair{0,1}, pair{0,1}); 
>    B(0,0) = 0;
> }
> ```

Thanks! Commit 5ab14d1 addresses this new issue. Now the slice of a `const legacyMatrix<T>` is a `legacyMatrix<const T>`. This was inspired in the way Eigen3 does it.

thijssteel (approved):

PR #214, opened 2023-04-04
Use concepts instead of general templates

thijssteel:
Design of this PR:

- Matrix concept (requires that (i,j) -> T access is available)
- Vector concept (requires that [i] -> T access is available)

Some struggles:
- Defining an arithmetic type doesn't immediately work, might be because Eigen::half doesn't satisfy std::is_arithmetic, not sure.
- Do we also define concepts for side_t, trans_t, ...? If so, then where, together with the other concepts or in the file where they are defined?

Are the names: `AbstractMatrix` and `AbstractVector` ok?


thijssteel:
A major disadvantage of `is_arithmetic`: `std::is_arithmetic_v<std::complex<float>>` is false. I.e. a complex number is not considered a number. I can't even find and standard type traits to check if a number is complex.

weslleyspereira:
> ex number is not considered a number. I can't even find and standard type traits to check if a number is complex.

Does `tlapack::is_complex<T>` help?

weslleyspereira:
Also, I want go back in my support for the names `AbstractMatrix` and `AbstractVector`. They are macros in header files, which means they do not belong to the namespace `tlapack`. So, we should change them to a name that are unique for \<T\>LAPACK. So far, we have been using uppercase `TLAPACK_` for macros in the library.  I suggest either `tlapack_matrix` or `TLAPACK_MATRIX` instead of `AbstractMatrix`.

weslleyspereira:
> `StrongZero` also does not satisfy it of course.

Please, try to create `template<> struct is_arithmetic<T> { enum { value = true }; };` for all the types needed. If you encounter problems, like lack of arithmetic operations or lack of the implementation of `numeric_limits` I can to help. All types we in \<T\>LAPACK should be arithmetic types.

weslleyspereira:
> > ex number is not considered a number. I can't even find and standard type traits to check if a number is complex.
> 
> Does `tlapack::is_complex<T>` help?

The rule could be: `is_arithmetic<real_type<T>>`, if indeed this is a valid rule for a concept.

thijssteel:
I'm currently leaning towards:
```c++
template <typename T>
concept NumberConcept = requires(T a, T b)
{
    a + b;
    a* b;
};
```
That way, we don't need to extend the trait for new types

weslleyspereira:
> I'm currently leaning towards:
> 
> ```c++
> template <typename T>
> concept NumberConcept = requires(T a, T b)
> {
>     a + b;
>     a* b;
> };
> ```
> 
> That way, we don't need to extend the trait for new types

That is fine for me too.

Mind that the trait is supposed to be extensible, so developers can mark new types as arithmetic. For instance, Catch2 uses this trait to better print the error message of `CHECK( a < b )`. If `a` and `b` are arithmetic, there is a change in the behavior of the macro as I understood.

weslleyspereira:
Hi @thijssteel. May I help you here? I can solve the conflicts if you want.

thijssteel:
> Hi @thijssteel. May I help you here? I can solve the conflicts if you want.

Yes, I don't really have time to work on <T>LAPACK right now. Feel free to make this PR your own.

weslleyspereira:
This PR is should be complete in its current status.

Concepts:

```c++
    /// Macro for tlapack::concepts::Matrix compatible with C++17.
    #define TLAPACK_MATRIX tlapack::concepts::Matrix

    /// Macro for tlapack::concepts::SliceableMatrix compatible with C++17.
    #define TLAPACK_SMATRIX tlapack::concepts::SliceableMatrix

    /// Macro for tlapack::concepts::Vector compatible with C++17.
    #define TLAPACK_VECTOR tlapack::concepts::Vector

    /// Macro for tlapack::concepts::SliceableVector compatible with C++17.
    #define TLAPACK_SVECTOR tlapack::concepts::SliceableVector

    /// Macro for tlapack::concepts::Scalar compatible with C++17.
    #define TLAPACK_SCALAR tlapack::concepts::Scalar

    /// Macro for tlapack::concepts::Real compatible with C++17.
    #define TLAPACK_REAL tlapack::concepts::Real

    /// Macro for tlapack::concepts::Complex compatible with C++17.
    #define TLAPACK_COMPLEX tlapack::concepts::Complex

    /// Macro for tlapack::concepts::Index compatible with C++17.
    #define TLAPACK_INDEX tlapack::concepts::Index

    /// Macro for tlapack::concepts::Side compatible with C++17.
    #define TLAPACK_SIDE tlapack::concepts::Side

    /// Macro for tlapack::concepts::Direction compatible with C++17.
    #define TLAPACK_DIRECTION tlapack::concepts::Direction

    /// Macro for tlapack::concepts::Op compatible with C++17.
    #define TLAPACK_OP tlapack::concepts::Op

    /// Macro for tlapack::concepts::StoreV compatible with C++17.
    #define TLAPACK_STOREV tlapack::concepts::StoreV

    /// Macro for tlapack::concepts::Norm compatible with C++17.
    #define TLAPACK_NORM tlapack::concepts::Norm

    /// Macro for tlapack::concepts::Uplo compatible with C++17.
    #define TLAPACK_UPLO tlapack::concepts::Uplo

    /// Macro for tlapack::concepts::Diag compatible with C++17.
    #define TLAPACK_DIAG tlapack::concepts::Diag
```

Other changes:
- CMake now requires C++17 and uses C++20 if available.
- New group `concepts` in the documentation. The documentation was improved in the process.
- Some types needed additional functions so that they satisfy the concepts specifications. Those functions were added.

Clean up:
- `Eigen::half` and `Eigen::Matrix` are in separate files now. They can be included separately.
- `MatrixAccessPolicy` was completely removed. Now `Uplo` takes a couple more tags.
- `is_arithmetic` was removed. There is a concept `Arithmetic` that is more clear in what we consider an arithmetic type.
- Multi-entry `min` and `max` were rarely used, so removed to clean namespace tlapack.
- `abstractArray.hpp` was removed. The concepts describe all we need from matrices and vectors.
- Name "interface" was removed from the documentation to avoid confusion with notion of "abstract class" in C++. We should avoid using the term "abstract" as well.

PR #221, opened 2023-04-25
Extends real_type<> for matrices

### LINKED ISSUE #219: Real matrix type
To my knowledge, there is no way to turn a matrix type into a real matrix type. Something like:
```c++
using real_matrix_t = real_t<matrix_t>;
```
Does this require extending all the plugins or are there better ways?

weslleyspereira:
Closes #219.

This PR:
- Creates `tlapack::is_real<T>` in addition to the `tlapack::is_complex<T>`. Why? We now may have real types, complex types and non-real non-complex types. Before, we were assuming everything that wasn't complex, was a real type. Moreover, matrices and vectors can now be identified as real or complex.
- Creates `tlapack::is_arithmetic<T>` that helps to separate scalar types from matrices and vectors.
- Updates the plugins so that they reflect the changes.
- Removed a couple of unused variables.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -34,7 +34,7 @@ permissions:
 
 env:
   CFLAGS: "-Wall -pedantic -Wno-unused-variable -Wno-unused-function"
-  CXXFLAGS: "-Wall -pedantic -Wno-unused-variable -Wno-unused-function"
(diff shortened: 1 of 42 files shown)


weslleyspereira:
I know where the problem is. I will solve tomorrow :)

PR #230, opened 2023-07-07
Remove dependency legacy arrays

weslleyspereira:
- There was still some dependency on LegacyMatrix, LegacyVector and LegacyBandedMatrix. I moved them out of `tlapack/base`. Their names now are following the CamelCase style, as is usual for class names in C++. I will update the naming conventions on the Contributing guidelines


weslleyspereira:
I have just updated the original message. The conventions now should cover most of the use cases in the library.

PR #231, opened 2023-07-13
Improvements in the workspace

weslleyspereira:
- [x] larfb and larf have additional interface with matrix W
- [ ] The optional workspace now is always column-major matrix. Reason: Simplify the code.

### DIFF
--- a/include/tlapack/base/workspace.hpp
+++ b/include/tlapack/base/workspace.hpp
@@ -183,17 +183,21 @@ struct WorkInfo {
         const size_t s1 = workinfo.size();
         const size_t s = size();
 
-        // Check if the current sizes do not cover the sizes from workinfo
-        if (!((m >= s1 && n >= 1) || (m >= 1 && n >= s1) ||
-              (m >= m1 && n >= n1))) {
+        // Check if the current sizes cover the sizes from workinfo
+        if ((m >= s1 && n >= 1) || (m >= 1 && n >= s1) ||
+            (m >= m1 && n >= n1)) {
+            // Nothing to do
+            return;
+        }
+        else {
             // Check if the sizes from workinfo cover the current sizes
             if ((m1 >= s && n1 >= 1) || (m1 >= 1 && n1 >= s) ||
                 (m1 >= m && n1 >= n)) {
+                // Use the sizes from workinfo
                 m = m1;
                 n = n1;
             }
-            else  // Incompatible workinfo. Use contiguous space the maximum
-                  // size
+            else  // Otherwise, use contiguous space with the maximum size
             {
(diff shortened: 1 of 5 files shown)


weslleyspereira:
The workspace is optional. Then, one could think in the following strategy 1:

1. If the workspace size is zero, then constructs a workspace matrix/vector that has the required sizes (workspace query) and whose type is the best type in respect to the matrices on input. For instance, if the input is `Eigen::MatrixXd` (double) and the routine requires a single workspace matrix of double values with size m-by-n, then we just use `Eigen::MatrixXd(m,n)`.
2. If the workspace size is greater than zero, then use the space previously allocated. In this case, if the input is `Eigen::MatrixXd` (double) and the routine requires a single workspace matrix of double values with size m-by-n, we could use `Eigen::Map<Eigen::MatrixXd>` as the workspace matrix. This class represents matrices that do not own their pointers of data.

C++ is strongly typed. Since workspace matrices (1) and (2) have different types, they should be stored into different variables. Each code would need to have two branches, one for each type of matrix. Alternatively, we could use opaque pointers `void*` to those matrices objects and let the correct implementation to be chosen at runtime. The alternative introduces a construct that we haven't being using in the library so far. Ginkgo uses such pointers, so we could look there if we want to go this way.

Strategy 2: "Well, let's change strategy 1 and have (1) and (2) both returning the same type". That is possible under limitations:
  a. `Eigen::Map<Eigen::MatrixXd>` would need to be the common type, because one of the data arrays is already allocated.
  b. `Eigen::Map<Eigen::MatrixXd>` does not owns the data, so, naturally, there must be a data array accessible from the scope where the matrix is defined, which can be accessed by the matrices.
  c. In case (2), we should at least allow for strided arrays of data, because this is closely related to how we define matrices in usual dense linear algebra. Therefore, we are limited to map the data array to strided layouts.
  d. What if I have an array of 
  
OBS: The second strategy I described is pretty much what we have in current code. We use an object of type `Workspace` to store information about the strided array of data, with two strides, one of them equal to 1. This can be mapped to column- and row-major matrices, strided vectors, and banded matrices in LAPACK.

One last limitation of second strategy: if we have a m-by-n strided array of data with stride ldim (n pieces of m contiguous data). We can transform it to either a m-by-n column-major matrix or a n-by-m column-major matrix. Good! So this creates opportunities of customization. However, it also creates some difficulty since the layout is usually defined at compile time. If one chooses to transform the data array into a column-major matrix, it will be one implementation. The specialization into row-major matrices lead to a second implementation of the routine. We end up in a problem similar to what we had in strategy 1.

Because of the latter, I am proposing to use a single data layout, say, column-major.

PR #239, opened 2023-08-04
Separate column-major specific implementation of the norm routines

weslleyspereira:
LANGE and the other norm routines currently have 2 interfaces. One of those interfaces use a workspace that allows to go over the matrix column by column in one pass.

So, here is my question: should we provide such a column-major specialization for LANGE and cia? If so, what is the best way to do that. Let me dive into this subject.

 

Forget for a minute the LANGE case and let us think more general. Suppose we have a routine "foo" that is in \<T\>LAPACK that does not need a workspace. Suppose multiple other routines call "foo" and does not ask for the workspace size of "foo", because they know "foo" does not have a workspace. Now, someone come with an algorithm that does the same task as "foo" but needs a workspace for that. What should we do? How does our design would cover this case?

 

We could think: let us assume that every routine in \<T\>LAPACK has a workspace, and then we always need to query before use it. Unfortunately, there is a problem with that strategy. The problem is that we don't know how many workspaces are needed and their types. So, should we assume that all routines will always have an integer, a float-precision and a complex workspace? What about boolean? That is too much, since most routines will have only a single workspace. It would make things complicated for users and developers.

 

My idea: a given routine will have a number 'x' of workspaces of pre-determined entry types. We state that in the documentation, so users and developers must be prepared to call workspace queries. If someone comes up with a variant of the same routine that uses a workspace, that variant should use another name. Is that reasonable?

 

Now, the case of LANGE, LANSY, LANHE and LANTR. I think we should state that LANGE is a routine that uses no workspace. The implementation of LANGE for column-major is a variant that must use another name.

-----

TO-DO:
- [x] Adds explanation in the documentation about how the workspaces should be documented.


weslleyspereira:
Explanation about how workspaces should be documented was already added at the end of https://github.com/tlapack/tlapack/wiki/Workspace#top-level-design

PR #245, opened 2023-08-16
Establish rules for using `inline` and `constexpr`

weslleyspereira:
# Use inline on functions that may appear on more than one translation unit.

The inline specifier was used also for performance in the library. However, compilers make the final decision about inlining or not a function. See the explanation in
https://en.cppreference.com/w/cpp/language/inline. The use of the inline specifier may change the priority for inlining a function by the compiler. It is worth noticing that forcing the inline may lead to large executables, which is specially bad when the library is already based on templates.

This commit puts the responsibility of inlining for performance in the compiler side. Future commits may establish better rules for inline functions in the library.


weslleyspereira:
Use `constexpr` specifier for:
- Utility functions like `safe_max()`, `abs1()` and `WorkInfo::minMax()`.
- Non-recursive workspace queries.

This commit also improves the definition of some local variables,
turning them into const references when it makes sense.

Issue #194, opened 2023-02-23
debian/testing: build fails


fghoussen:
Reporting here that build fails on debian/testing:
```
>> cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux bookworm/sid"
NAME="Debian GNU/Linux"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

~/Workspaces/tlapack/build >> cmake ..

~/Workspaces/tlapack/build >> cmake -LA .
-- <T>LAPACK version 0.1.1
-- Submodule update
-- ClangFormat found: /usr/bin/clang-format (version )
-- Could NOT find MPFR (missing: MPFR_INCLUDES MPFR_LIBRARIES MPFR_VERSION_OK) (Required is at least version "2.3.1")
-- Could NOT find GMP (missing: GMP_INCLUDES GMP_LIBRARIES) (Required is at least version "4.2.1")
-- Could NOT find MPFR (missing: MPFR_INCLUDES MPFR_LIBRARIES MPFR_VERSION_OK) (Required is at least version "2.3.1")
-- Could NOT find GMP (missing: GMP_INCLUDES GMP_LIBRARIES) (Required is at least version "4.2.1")
-- Looking for LAPACKE_dgeqrf
-- Looking for LAPACKE_dgeqrf - not found
-- testBLAS version 0.1.2
-- Configuring done
-- Generating done
-- Build files have been written to: /home/fhoussen/Workspaces/tlapack/build
-- Cache values
BLAS_Accelerate_LIBRARY:FILEPATH=BLAS_Accelerate_LIBRARY-NOTFOUND
BLAS_acml_LIBRARY:FILEPATH=BLAS_acml_LIBRARY-NOTFOUND
BLAS_acml_mp_LIBRARY:FILEPATH=BLAS_acml_mp_LIBRARY-NOTFOUND
BLAS_armpl_lp64_LIBRARY:FILEPATH=BLAS_armpl_lp64_LIBRARY-NOTFOUND
BLAS_blas_LIBRARY:FILEPATH=/usr/lib/x86_64-linux-gnu/libblas.so
BLAS_blis_LIBRARY:FILEPATH=BLAS_blis_LIBRARY-NOTFOUND
[...]

PR #256, opened 2023-08-24
This commit organizes the dependencies to external projects

weslleyspereira:
Moreover:
- It automatically downloads the test directories from BLAS++ and LAPACK++.
- Creates a section in the README.md to explain the dependencies.


weslleyspereira:
Related to #194

PR #269, opened 2023-09-07
add unblocked hessenberg-triangular reduction

thijssteel:


weslleyspereira (approved):
I think this is ready for merging, right? Thanks!

PR #271, opened 2023-09-08
More compilation issues

thijssteel:


thijssteel:
@weslleyspereira was it on purpose that the ci always uses c++17? If so i can revert that change.

weslleyspereira:
> @weslleyspereira was it on purpose that the ci always uses c++17? If so i can revert that change.

It was intentional. I had some issues that only appeared using C++17. However, I agree with your changes. Maybe we want to keep at least one job in the CI using C++17.

thijssteel:
> Maybe we want to keep at least one job in the CI using C++17.

In this PR, the examples are still compiled using c++17, so should be fine.

PR #276, opened 2023-09-12
Adds test for norms

weslleyspereira:
Adds a few tests for the norms


weslleyspereira:
While writing the tests, I found out the assumptions in `lassq`  for `scale` was undesirable. That is because the norm routines needs to use `lassq` with `scale` out of the range we were assuming to have. This was generating errors in the norm computations.

Therefore, this PR also removes the assumptions: 

```c++
 * If (scale * sqrt( sumsq )) > tbig on entry then
 *    we require:   scale >= sqrt( TINY*EPS ) / sbig   on entry,
 * and if 0 < (scale * sqrt( sumsq )) < tsml on entry then
 *    we require:   scale <= sqrt( HUGE ) / ssml       on entry,
```

weslleyspereira:
I will soon merge this PR since it solves a bug in the norm computation.

PR #278, opened 2023-09-13
Blocked ht reduction

thijssteel:
Adds blocked Hessenberg-triangular reduction, based on dgghd3 in LAPACK.

Some changes w.r.t. the LAPACK code:

- LAPACK stores the rotations in the empty space underneath the matrices. I decided to store them in a separate workspace. The extra required space is not that large and this change makes it easier to parallelize it later (we can parallellize the accumulation)
- LAPACK accumulates the left rotations while reducing the panel (they have to because they don't have enough space to store all the rotations). I only accumulate the rotations after the panel has been reduced. Note, LAPACK also uses the accumulated rotations to update the panel during the reduction, but i don't believe this is important for performance.
- I made an extra helper function rot_sequence which efficiently applies a sequence of rotations to a matrix. This efficiency is achieved by changing the loop orders depending on row/column layout and doing some loop unrolling
- I made an extra helper function hessenberg_rq to extract some of the functionality. I also changed how this routine works a little bit. LAPACK cheats a little by applying a permuted rotation (i.e. they swap the columns), but this made it incompatible with rot_sequence, fixing that is a small change (basically just changing the sign of the sine) In the future, this could also make it easier to make this routine faster since it is a big performance bottleneck.

The version in this PR is slower than the unblocked HT reduction, I will investigate further so this is a draft for now.


thijssteel:
@weslleyspereira I just did a quick experiment, and I can pretty easily explain the poor performance, i was not using optimized blas (i put a print statement in the unoptimized blas and it does print).

weslleyspereira:
> @weslleyspereira I just did a quick experiment, and I can pretty easily explain the poor performance, i was not using optimized blas (i put a print statement in the unoptimized blas and it does print).

Great that you found! Maybe we could think of a non-intrusive strategy to reporting the usage of non-optimized routines when `TLAPACK_USE_LAPACKPP` is defined. It would be good to have a tracker for this behavior. It is not the first time we call non-optimized routines without noticing.

thijssteel:
Latest timings:

Testing gghd3 (TLAPACK)
n = 250 time = 0.0314619 s
n = 500 time = 0.19326 s
n = 1000 time = 1.01531 s
n = 2000 time = 7.62389 s
Testing lapack 
n = 250 time = 0.0196767 s
n = 500 time = 0.110596 s
n = 1000 time = 0.67312 s
n = 2000 time = 4.83685 s
Testing gghrd (TLAPACK)
n = 250 time = 0.0223004 s
n = 500 time = 0.162464 s
n = 1000 time = 1.36882 s
n = 2000 time = 14.682 s

Achieved with oneapi 2023 on an i7-1255U CPU, gcc compilers using -O3 -DNDEBUG -DTLAPACK_NDEBUG. Clearly, my code is a lot slower than MKL, but at this point i am completely out of ideas.

I also forced the use of optimized blas by commenting out the unoptimized blas and removing the enable_if_allows_opt_blas<> clause from the lapackpp call.

I'll mark this for review now and maybe sort it out later.

weslleyspereira (approved):
Great PR! I also think we can merge it now. We can always revisit the code later to see what is causing the performance loss.

langou:
WRT timing, @thijssteel: when you write "lapack" do you mean Intel MKL gghd3  or do you mean Netlib gghd3 with Intel MKL BLAS? I think you mean Intel MKL gghd3. Maybe it is worth comparing to Netlib gghd3 with Intel MKL BLAS too. I am assuming that the performance of Intel MKL gghd3  and Netlib gghd3 with Intel MKL BLAS are more or less the same but I do not know for sure and it could be good to check.

thijssteel:
Intel MKL gghd3 is prettyslow (they appear to set nb to 1), i compiled netlib gghd3, set nb to 32 and linked with MKL blas

PR #281, opened 2023-09-19
First changes to test the idea

weslleyspereira:


thijssteel:
My first thought was 'reshape is not a proper name anymore', but i don't have any better suggestions.

This addresses many of my concerns about the previous design. It is much easier to manage the workspace now.

Some concerns:

**How do we handle different workspace types?**

If this design allows a workspace like std::vector of bytes, does that mean that all workspace should be reshapeable into different types as well? Or would we still require different workspace for real and complex?

**Some concerns about reshaping non-contiguous matrices**

What if I have an (n x k) workspace and I slice of a size 3 vector, what will happen to the remainder?
- will it be reshaped into (n-1 x k) or into (n x k-1), will this depend on the layout? What happens to the k-3 / n - 3 space that remains?
- if the original workspace was contiguous, i assume it may be reshaped into a vector of size 3 and a vector of size k*n-3.

The combination of those two behaviors is dangerous. We usually test with contiguous workspace, so we may miss some bugs. The same goes for the different plugins. We'd basically require the plugins to be fully consistent in how they slice up matrices, but they may want to make different decisions for optimal performance, making sure the matrix is internally contiguous, alignments...
 
**Why do we need this?**

I suggested a simpler design, without support for discontiguous workspace. I have yet to see a concrete case where it cannot easily be replaced with a matrix. When we discussed it over zoom, you worried that such a change would propagate to all the calling routines, but I believe it shouldn't really matter to the calling routines, they can continue to provide a matrix as they do now. I may be wrong, I just want to be convinced we need it before we settle on this design.

weslleyspereira:
I'll comment on the first 2 topics first.

> Some concerns:
> 
> **How do we handle different workspace types?**
> 
> If this design allows a workspace like std::vector of bytes, does that mean that all workspace should be reshapeable into different types as well? Or would we still require different workspace for real and complex?

No for the first question. Let me give you an example.

Suppose we implement a routine `svd_work(matrix_t& A, work_t& work)`, and this routine expects a workspace of bytes. Inside the routine, we would call `reshape(work, m, n)` to reshape our workspace to some other matrix (or vector if we drop `m`). However, since `work` has no information about type or layout, we would need more elements in this call. Something like: `reshape<SomeMatrixType>(work, m, n)`. So, those workspaces of bytes can be reshaped to matrices and vectors with any data type and any layout. The routine `svd_work()` would need to come up with the suitable `SomeMatrixType`.

The way it is implemented now, a routine that expects to receive a workspace of type `double` cannot handle a workspace of bytes.

> **Some concerns about reshaping non-contiguous matrices**
> 
> What if I have an (n x k) workspace and I slice of a size 3 vector, what will happen to the remainder?

In this case, I opt to raising an exception instead of doing anything else. I don't think we need to deal with non-contiguous workspaces in such a fine level. Non-contiguous workspaces are useful on internal calls for some routines and those routines should deal with remainders in my opinion. This way, each plugin may make the decisions based on their optimal way in terms of performance.

weslleyspereira:
> **Why do we need this?**
> 
> I suggested a simpler design, without support for discontiguous workspace. I have yet to see a concrete case where it cannot easily be replaced with a matrix. When we discussed it over zoom, you worried that such a change would propagate to all the calling routines, but I believe it shouldn't really matter to the calling routines, they can continue to provide a matrix as they do now. I may be wrong, I just want to be convinced we need it before we settle on this design.

I don't have a concrete example on the practical usage of non-contiguous workspaces aside for the ones on internal calls. I think we could ask and see what we get. I don't think it is too bad to be prepared for that. You are correct about the testing coverage, and I think we should add tests for all `reshape()` functions. I don't think we need to use non-contiguous workspaces on the test suite if we have tests for the `reshape()` functions in the plugins.

That said, I do see a potential benefit of the current strategy in comparison to the usage of contiguous vectors with raw data. For instance, we could have thought that `unmhr()` is a legit routine that deserves its contiguous workspace with raw data in `unmhr_work()`. However, now we have `aggresive_early_deflation()` which calls `unmhr()` internally using parts of the input matrix A. So, in this case, what should be the correct way to define the workspace of `unmhr()` (or `unmhr_work()` if you want)? Wouldn't it be possible that we have a routine in the future that needs `unmqr()` internally possibly using parts of the input as workspace? I just want don't want to close this door yet.

There is something else besides the battle: contiguous x non-contiguous. Matrices have layout and maybe a grid on top of that (when we use StarPU for instance). My question is: who should hold the responsibility of choosing the correct layout and other properties of a workspace? I don't have an answer.
1. The current design allows the user to decide.
2. If we a vector with raw data, the responsibility is of each routine that will give shape to the workspace.

Option 1 is more flexible. If the user wants to provide a workspace, it should tell us the type, layout and other properties. Maybe it is useful to have a row-major workspace for col-major input matrices in a specific routine. I don't know. If the user does not want to bother with defining workspaces, they will use the interface that allocates everything for them.

Option 2 is more ease to use on the user side. The user will either provide nothing as workspace or provide some piece of raw data they have. The developer would have the flexibility/job to decide what is the best thing to do with that raw data. In the end, we may end up on an automated way of deciding the layout of each workspace, using, for instance, `matrix_type<matrixA_t,matrixB_t>`. Everything would work and be simple that way.

This is the way I see things.

weslleyspereira:
With any design we choose, it could be useful to go a bit further and define interfaces that receives each workspace matrix and vector as a separate argument. For instance:
`int geqrf_work(A_t& A, tau_t& tau, matrixT_t& matrixT, W_t& W, const GeqrfOpts& opts = {})`
instead of
`int geqrf_work(A_t& A, tau_t& tau, work_t& work, const GeqrfOpts& opts = {})`.

Those interfaces could be internal to the library.

thijssteel:
> That said, I do see a potential benefit of the current strategy in comparison to the usage of contiguous vectors with raw data. For instance, we could have thought that unmhr() is a legit routine that deserves its contiguous workspace with raw data in unmhr_work(). However, now we have aggresive_early_deflation() which calls unmhr() internally using parts of the input matrix A. So, in this case, what should be the correct way to define the workspace of unmhr() (or unmhr_work() if you want)? Wouldn't it be possible that we have a routine in the future that needs unmqr() internally possibly using parts of the input as workspace? I just want don't want to close this door yet.

unmhr is not an internal/specialized routine and absolute should receive its workspace in the normal way. But that is not a problem. Yes, AED currently calls unmhr using parts of the matrix, but it also has a normal workspace that it can use. Using the parts below the matrix is more important for the gemm calls and the deflation window. In fact, looking at it now, i think using parts of the matrix instead of the provided workspace was a mistake.

> What if I have an (n x k) workspace and I slice of a size 3 vector, what will happen to the remainder? In this case, I opt to raising an exception instead of doing anything else. I don't think we need to deal with non-contiguous workspaces in such a fine level. Non-contiguous workspaces are useful on internal calls for some routines and those routines should deal with remainders in my opinion. This way, each plugin may make the decisions based on their optimal way in terms of performance.

I really don't like that idea. How is a routine supposed to deal with a remainder exactly? If it is the responsibility of the routine, then I think we end up with slices again, not a reshape. And this is not the case for a few routines where we need discontiguous workspace. We need to deal with this for every single routine.

weslleyspereira:
I will merge this PR since it presents some improvements to the current workspace design. We will for sure continue discussion on the topic.

weslleyspereira:
Current workspace design is described in https://github.com/tlapack/tlapack/wiki/Workspace.

PR #284, opened 2023-09-20
Update minimum CMake requirements to CMake 3.5

weslleyspereira:
Update minimum CMake requirements to CMake 3.5 as compatibility with CMake < 3.5 will be removed from a future version of CMake.


weslleyspereira:
Another quick update that needs no review. Just merging it to avoid annoying messages from in the CMake.

PR #289, opened 2023-09-27
Qz

thijssteel:
Implementation of implicit QZ

small parts of the implementation depend of my SVD pr (the normalization of 2x2 blocks). I'll fix those later.

An important part is also the calculation of the generalized eigenvalues of a 2x2 pencil. I didn't really feel like implementing that at the moment so i'll leave that for a later PR.


thijssteel:
This should be ready to merge now

weslleyspereira (approved):
Awesome!!

PR #290, opened 2023-09-27
Fix annoying warnings related to __float128 in GNU compilers with pedantic flag

weslleyspereira:


weslleyspereira:
Merging it since there is no much brain involved in those changes.

PR #291, opened 2023-09-27
Fixes windows build

weslleyspereira:
Another attempt to fix the most recent bugs in the CI for Windows.


weslleyspereira:
Changes in this PR:
- Using windows-2019 instead of windows-2022 (windows-latest): In windows-2022, the GCC compiler gives an error `0xc0000139`, which means "missing library". I couldn't figure out what is the exact problem there. I have also tried to use clang on windows-2022, but it gives me an incompatibility error with a non-standard header installed on windows.  
- Do not test libquadmath on Windows: the function `quadmath_snprintf()` seems to have a bug or it does not exist. I don't know for sure.

PR #292, opened 2023-09-28
add debugutils for eigen::half and vectors

thijssteel:
I often find myself in a situation where only the Eigen::Half tests are failing. Which is annoying because there are no pretty printers for Eigen::Half in gdb.

This PR extends the debugutils for visualizing matrix to vectors and adds functions for Eigen::Half.


weslleyspereira (approved):
Thanks!

weslleyspereira:
Actually, I see some problems with the CI. I think merge as soon as the CI likes the new code.

PR #294, opened 2023-09-29
Fix integration with optimized BLAS

### LINKED ISSUE #287: Use of optimized blas
I just did a quick experiment to check if optimized blas is being used.

```c++

    using T = double;
    using matrix_t =
        tlapack::LegacyMatrix<T, std::size_t, tlapack::Layout::ColMajor>;

    std::cout << "<T>: " << tlapack::allow_optblas<T> << std::endl;
    std::cout << "<matrix_t>: " << tlapack::allow_optblas<matrix_t> << std::endl;
```

For double, optimized blas is being used, but for legacymatrix, the result is false. I think there might be a problem with the plugin.

weslleyspereira:
This PR fixes a bug in the integration with optimized BLAS. It also make the CI tests more strict so that those bugs can be better tracked.

Closes #287.


weslleyspereira:
By making tests more strict, I found out an issue with our proof-of-concept Fortran interface. The interface that was not using the module file had a bug that appeared on Windows only. Now, the bugs were fixed and all examples use the module file.

PR #296, opened 2023-10-02
Try to reduce compile time of tests

weslleyspereira:
- Removes undesirable `link_libraries()` from CMake.


weslleyspereira:
### Analyzing the build time

I am using two measurements:
1. Time that each target takes to be built. I use the Ninja build system, so I have a report of the build times in the file `.ninja_log`.
2. Clang time tracer. I use clang++ with the compiler flag `-ftime-trace`, which collects more detailed information about build times in one .json per build target.

There is a lot of information in the .json files, so it is interesting to use another software to extract the information we want. You can open each file in the Google Chrome tracing tool. I didn't like the way the results are displayed, and having multiple .json files make it harder to analyze. So, I found a nice tool that does the job: https://github.com/aras-p/ClangBuildAnalyzer.

### First attempt: replace `TEMPLATE_TEST_CASE` by `TEMPLATE_LIST_TEST_CASE`.

- This change solely does not seem to affect the build time.
- I did noticed a slightly modification in the test executable, but it is something really small. I don't think it is worth doing the change just because of that.

weslleyspereira:
### Second attempt: Declare templates classes explicitly

This change does reduce the build time of the targets. Basically, [b5c6007](https://github.com/tlapack/tlapack/pull/296/commits/b5c60076e01a5f415653d3e9a58baa1e7d1ec9fc) explicitly declares the matrix types we use for the tests, and then those matrices do not need to be instantiated for each build target. They are instantiated in the `testutils` library.

thijssteel:
I don't fully understand how this works. But i am definitely in favor of its goals.

thijssteel:
On a related note. Is it worth doing some caching in the ci? I see that a lot of time is spent reinstalling the necessary libraries each time.

weslleyspereira:
Thanks! Yes, there is a page about that: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

I looked into it a while ago, and I think this was indeed possible in our case. I will have another look.

weslleyspereira:
I tried to do some caching, as suggested, using [cache-apt-pkgs-action](https://github.com/marketplace/actions/cache-apt-packages). It didn't work great. I found a problem:

```
 22:10:50.712   * Caching intel-oneapi-icc-eclipse-plugin-cpp-2023.2.1 to /home/runner/cache-apt-pkgs/intel-oneapi-icc-eclipse-plugin-cpp-2023.2.1=2023.2.1-16.tar...
tar: Cowardly refusing to create an empty archive
```

I will let the caching for another PR.

PR #307, opened 2023-10-20
Multishift qz

thijssteel:
Note, i just wanted to track my own changes, this is still a draft.

Changes in this PR:
- implement inverse householder reflector of order 3.
- Use inverse householder in unblocked QZ. This resulted in a surprizing amount of speedup. Almost 30%
- implement schur swap (this has a significant change w.r.t. lapack, because i use 2x2 svd and givens instead of QR), should be significantly more accurate, but i need to do some benchmarks to see if it is also fast.
- implement multishift sweep

TODO:
- implement schur move
- implement AED
- implement full multishift QZ

### DIFF
--- a/include/tlapack/lapack/generalized_schur_swap.hpp
+++ b/include/tlapack/lapack/generalized_schur_swap.hpp
@@ -661,7 +661,56 @@ int generalized_schur_swap(bool want_q,
         }
     }
     if (n1 == 2) {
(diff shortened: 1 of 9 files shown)


weslleyspereira:
The merge was to resolve a conflict with a CMakeLists.txt

weslleyspereira (approved):

PR #311, opened 2023-10-25
Adds CreateStaticFunctor for allocation of arrays on the stack

### LINKED ISSUE #288: small fixed size matrices/vectors on the stack
Creating a new vector, currently requires std::vector to be used in the create functor.

Some routines require a small, fixed size, amount of workspace that is normally allocated on the stack, but std::vector is (i think) always on the heap.

For example, my QR and QZ implementations currently use this:
https://github.com/tlapack/tlapack/blob/master/include/tlapack/lapack/lahqr.hpp#L295

It would be nice to be able to wrap a fixed size array to interpret it as a vector.

weslleyspereira:
Closes #288.


weslleyspereira:
This should be ready to merge.

@thijssteel, could you please take a look at [lasy2.hpp](https://github.com/tlapack/tlapack/pull/311/files#diff-328f922179023d8abb2dd5bd4061346b4803b6bd616c5e9d696e8e9eb1a5afb7)? I needed to change the template arguments a bit. Thanks.

thijssteel:
looks good

PR #319, opened 2023-11-03
Removes the need of TLAPACK_PREFERRED_MATRIX_* macros

weslleyspereira:
This PR cleans up the macros that deduce the common type of a collection of matrices and/or vectors. I remove the `TLAPACK_PREFERRED_MATRIX_*` macros which, in my opinion, lead to more intuitive code.

TO DO:
- [x] Explain `matrix_type` and `vector_type` in the concepts.
- [x] Say a few words about that in the Wiki.


weslleyspereira:
Some of the benefits of this PR:

1. `std::vector` is a pivot type that will work with Eigen, mdspan and tlapack legacy plugins. Before, the slice of an `std::vector` was returning a `tlapack::LegacyVector`, which created a dependency on legacy data types for all plugins interacting with `std::vector`. With this PR, the slice of an `std::vector` returns a `std::span`.
2. Mechanism for using multiple plugins as described at https://github.com/tlapack/tlapack/wiki/Common-type-deductions#deducing-common-type-with-multiple-plugins

PR #328, opened 2023-11-17
Fix ifx bug in the CI

weslleyspereira:
See the bug in https://github.com/tlapack/tlapack/actions/runs/6868061223/job/18679777543.

This PR fixes it by using gfortran instead.


weslleyspereira:
I found out the problem: the oneAPI was not installing any fortran compiler at `/opt/intel/oneapi/compiler/latest/linux/bin/` when it should. I am not sure why this is happening. Let's just use GCC compilers for now.

PR #348, opened 2024-01-12
Symmetrix eigenvalues qr algorithm

thijssteel:
This PR adds the QR/QL algorithm for symmetric eigenvalue problems.


thijssteel:
I don't think the mkl failure is due to this pr. Should be good to merge.

PR #352, opened 2024-01-15
implement wavefront algorithm for sequence of rotations

thijssteel:
In the paper: "Restructuring the Tridiagonal and Bidiagonal QR Algorithms for Performance" F. G. Van Zee, R. A. Van de Geijn, G. Quintana-Orti

The authors describe a wavefront algorithm to efficiently apply givens rotations. If we can apply multiple sequences of rotations at a similar flop rate as a gemm call, we can greatly speed up things like singular value and hermitian eigenvalue problems if using the implicit QR algorithm.

Some speed comparisons are below:
- CPU: i7-1255U
- Number of columns in C and S (number of sequences of rotations to apply): 40
- Matrix A is square and of size n
- Timing is average of 8 runs with 2 warmup runs

From the left side (inherently less efficient because matrices are column major)
| n        | blocked(s) | unblocked(s) |
|------    |-----------|--------|
| 256    | 0.0037          |   0.0078     |
| 512    | 0.013          |   0.026     |
| 1024  | 0.048          | 0.107       |
| 2048  | 0.20          |  0.413      |
| 4096  | 0.78           |  1.63      |

Speedup is around 2

From the right side
| n        | blocked(s) | unblocked(s) |
|------    |-----------|--------|
| 256    | 0.00050         |   0.0025     |
| 512    | 0.0026         |  0.0094     |
| 1024  |0.0093         | 0.031      |
| 2048  | 0.030         |  0.12      |
| 4096  |  0.12         |  0.48      |

Here, the speedup is around 4

I also added some openmp pragmas, since the blocked routine allows for a trivial parallelization of the column/row blocks. This only results in minor speedups of ~1.5

### DIFF
--- a/include/tlapack/lapack/rot_sequence3.hpp
+++ b/include/tlapack/lapack/rot_sequence3.hpp
@@ -337,19 +337,71 @@ void rot_sequence3(
                         }
                     }
                     // Pipeline phase
-                    for (idx_t j = l - 1; j < n - 1; ++j) {
-                        for (idx_t i = 0, g = j; i < l; ++i, --g) {
+                    for (idx_t j = l - 1; j + 1 < n - 1; j += 2) {
+                        for (idx_t i = 0, g = j; i + 1 < l; i += 2, g -= 2) {
                             for (idx_t i1 = ib; i1 < ib2; ++i1) {
+                                //
+                                // Apply first rotation
+                                //
+
+                                // A(i1,g) after first rotation
+                                T temp1 = C(g, i) * A(i1, g) +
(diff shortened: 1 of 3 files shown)


thijssteel:
This should be ready for review now. The code seems really long and complicated, but it's mostly just different cases of the same simple code.

The performance for the application from the left (for column major) is still quite bad, but i don't think this is something i can solve. I know the algorithm is implemented in BLIS (at least they say so in the paper) but i don't know BLIS that well and I couldn't find it.

thijssteel:
I'm going to merge this. I need it for my next PR.

PR #359, opened 2024-02-09
Redo rotsequence3

thijssteel:


thijssteel:
@weslleyspereira the MKL failures really have me stumped, they keep popping up and i can't solve it, do you know what is happening?

PR #442, opened 2025-03-25
Improve the random generation in the MatrixMarket

weslleyspereira:
- `rand_generator` is renamed as `PCG32`. This uses the algorithm PCG-XSH-RR from https://www.pcg-random.org/pdf/hmc-cs-2014-0905.pdf with implementation from https://github.com/imneme/pcg-cpp. The advantage is that the PCG is well studied and has better properties than the implementation currently in our library.
- `rand_helper()` now receives to parameters: a random number generator and a distribution. The default is the uniform distribution in [0,1).
- There is no `rand_helper()` using the `rand()` generator. This implementation depends on the compiler, so not a good choice.

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -37,7 +37,7 @@ jobs:
     - name: Run clang-format style check for C/C++ programs.
       uses: jidicula/clang-format-action@4726374d1aa3c6aecf132e5197e498979588ebc8 # v4.15.0
       with:
-        clang-format-version: '10'
+        clang-format-version: '11'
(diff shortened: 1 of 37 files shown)


weslleyspereira:
Additionally:

- Skip QZ and Sylvester solver tests for FP16 type. Maybe you would be interested in this, @thijssteel.
- Weaken some tolerances based on test results.
- Fix gnuquad plugin compatibility with non-Intel (I think it is ARM) architecture in the github runner for MacOS.
- Fix format issues and move to ClangFormat v11, since Clang v10 is currently more difficult to install (e.g., not in brew for macos).
- Changes in LADIV: See following message.

PR #453, opened 2025-06-05
Try mult llh and uhu

BrianDang03:
This contributes llh. We need uhu. Someone else with contribute uhu in include. I will finish the integration of uhu with testing then. Please consider for merge.

### DIFF
--- a/include/tlapack/lapack/mult_llh.hpp
+++ b/include/tlapack/lapack/mult_llh.hpp
@@ -15,8 +15,8 @@
 
 namespace tlapack {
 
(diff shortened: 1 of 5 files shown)


include/tlapack/lapack/mult_llh.hpp
```diff
@@ -0,0 +1,101 @@
+/// @file lu_mult.hpp
+/// @author Lindsay Slager, University of Colorado Denver, USA
```
langou, line 2: Please change author to Brian. 🤣

PR #458, opened 2025-06-09
Test gemmtr

luis-carlos-gutierrez:
Created a blas routine gemmtr and tested successfully.


test/src/test_gemmtr.cpp
```diff
@@ -0,0 +1,157 @@
...
+        lacpy(GENERAL, C0, C2);
+
+        // Defining real and complex scalars
+        T alpha, beta;
+
+        if constexpr (is_complex<T>) {
+            alpha = T(static_cast<float>(0xDEADBEEF),
+                      static_cast<float>(0xDEADBEEF));
+            beta = T(static_cast<float>(0xDEADBEEF),
+                     static_cast<float>(0xDEADBEEF));
```
langou, line 103: @luis-carlos-gutierrez and @kcunningturkey 
Can you revert to alpha and beta being more "normal" constants?
I think the initial
```cpp
            alpha = T(-2, 5);
            beta = T(-7, 4);
```
was better

weslleyspereira: Also, the CI is not happy with that part of the code the way it is now.

langou: @luis-carlos-gutierrez and @kcunningturkey: can you try something like the below?

```cpp
typedef real_type<T> real_t;
T alpha, beta;
if constexpr (is_complex<T>) {
                alpha = T(real_t(-2), real_t(5));
                beta = T(real_t(-7), real_t (4));
} else {
                alpha = T(-3);
                beta = T(8);
}
```

test/src/test_gemmtr.cpp
```diff
@@ -145,12 +143,12 @@ TEMPLATE_TEST_CASE("check for gemmtr multiplication",
...
                       tol * (abs1(alpha) * norma * normb + abs1(beta) * normc));
 
-                T sum = T(0);
+                real_t sum = 0;
                 for (idx_t i = 0; i < n; i++)  // Check strictly upper part
                     for (idx_t j = i + 1; j < n; j++)
                         sum += abs1(C1(i, j) - C0(i, j));
 
-                CHECK(sum == T(0));
+                CHECK(sum == 0);
```
langou: Ditto. Try with
```cpp
CHECK(sum == real_t(0));
```

include/tlapack/lapack/gemmtr.hpp
```diff
@@ -0,0 +1,319 @@
...
+          TLAPACK_MATRIX matrixB_t,
+          TLAPACK_MATRIX matrixC_t,
+          TLAPACK_SCALAR alpha_t,
+          TLAPACK_SCALAR beta_t,
+          class T = type_t<matrixC_t>,
+          disable_if_allow_optblas_t<pair<matrixA_t, T>,
+                                     pair<matrixB_t, T>,
+                                     pair<matrixC_t, T>,
+                                     pair<alpha_t, T>,
+                                     pair<beta_t, T> > = 0>
```
weslleyspereira, line 65: You can safely remove those lines. They only make sense for routines that are implemented by BLAS++ or LAPACK++. All tests should work without them

Issue #459, opened 2025-06-09
please remove this line of code


langou:
Hi @BrianDang03, 

Can you remove this line of code:
https://github.com/tlapack/tlapack/blob/fca2fc7bc631e2fa701ee6f2076c4a5dc2436ec7/test/src/test_potrf.cpp#L98
and that one too
https://github.com/tlapack/tlapack/blob/fca2fc7bc631e2fa701ee6f2076c4a5dc2436ec7/test/src/test_potrf.cpp#L85

Also I think these two lines of code can be safely removed since E is not used any longer
https://github.com/tlapack/tlapack/blob/fca2fc7bc631e2fa701ee6f2076c4a5dc2436ec7/test/src/test_potrf.cpp#L100-L101

Julien

PR #460, opened 2025-06-10
Update test_potrf.cpp

BrianDang03:
Removed comments and unnecessary E matrix for test_potrf.cpp.

#459 

https://github.com/tlapack/tlapack/issues/459#issue-3131719739


include/tlapack/lapack/potri.hpp
```diff
@@ -0,0 +1,88 @@
+/// @file potrf2.hpp Computes the Cholesky factorization of a Hermitian positive
+/// definite matrix A using the recursive algorithm.
+/// @author Weslley S Pereira, University of Colorado Denver, USA
```
langou, line 3: Please change the name of the author

include/tlapack/lapack/potri.hpp
```diff
@@ -0,0 +1,88 @@
+/// @file potrf2.hpp Computes the Cholesky factorization of a Hermitian positive
```
langou, line 1: potri.hpp, not potrf2.hpp

langou: And well change the comment, you can look at LAPACK dpotri.f to take inspiration.

test/src/test_potri.cpp
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        // C - A
+        for (idx_t j = 0; j < n; j++) {
+            for (idx_t i = 0; i < n; i++) {
+                C(i, j) -= I(i, j);
+            }
+        }
+
+        real_t error = lange(FROB_NORM, C) / normA * normAIn;
```
langou, line 157: I think you want 
```cpp
error = lange(FROB_NORM, C) / normA / normAIn;
```
here.

include/tlapack/lapack/potri.hpp
```diff
@@ -0,0 +1,88 @@
...
+void potri(Uplo uplo, matrix_t& A)
+{
+    tlapack_check(uplo == Uplo::Lower || uplo == Uplo::Upper);
+    tlapack_check(nrows(A) == ncols(A));
+
+    potrf2(uplo, A);
+
+    trtri_recursive(uplo, Diag::NonUnit, A);
+
+    lauum_recursive(uplo, A);
```
langou, line 83: This is code and the code looks very nice and readable. This is one comment for future but here we are using some specific variants of POTRF, TRTRI and LAUUM for POTRI and I am not sure what the design should be. First we do not have the high level function. So we would need an lauum.hpp that is generic and has an "options" argument and depending on options you call one of the variant. We do not have this. And then this options parameter should probably be coming as an argument from potri.hpp. Ditto for portf.hpp and lauum.hpp. All good for now, just a comment.

include/tlapack/lapack/potri.hpp
```diff
@@ -0,0 +1,88 @@
...
+ *      Define the behavior of Exception Handling.
+ *
+ * @return = 0: successful exit
+ * @return i, 0 < i <= n, if the leading minor of order i is not
+ *     positive definite, and the factorization could not be completed.
+ *
+ * @ingroup computational
+ */
+template <TLAPACK_SMATRIX matrix_t>
+void potri(Uplo uplo, matrix_t& A)
```
langou, line 74: So (1) here for uplo, I think the interface uses the "enum" type and I wonder whether we want to use constant expression. See how this is done in portf.hpp for example. (2) Also it could be a good to pass in "opts" even if we do not have any opts. (3) And I think we want potri() to return an int and not a void. 

Your interface would be
```cpp
template <TLAPACK_UPLO uplo_t, TLAPACK_MATRIX matrix_t>
int potri(uplo_t uplo, matrix_t& A, const PotrfOpts& opts = {})
```

For now, please do nothing and leave your interface as is. This is not a problem to me and ready to merge, We can change this later.

langou:
Comment. 

This PR#460 was intended to fix issue #459, but this PR was never merged. 

It was superseded by PR #466.

PR#466 fixes issue #459.

PR #484, opened 2025-07-07
Updated Docs and Implemented Quick sort for steqr

BrianDang03:


test/src/test_steqr.cpp
```diff
@@ -58,28 +58,34 @@ TEMPLATE_TEST_CASE("steqr is backward stable",
...
 
         laset(Uplo::General, zero, one, Q);
-        int err = steqr(true, d, e, Q);
+        int err = steqr(false, d1, e1, Q);
+        REQUIRE(err == 0);
+        err = steqr(true, d2, e2, Q);
         REQUIRE(err == 0);
 
-        // Check that singular values are sorted in ascending
+        // Check that eigenvalues are sorted in ascending
```
langou, line 85: Yes, for the first check, write something like:
"check that eigenvalues of steqr(false, . . ) and steqr( true, . . .) are bit-by-bit identical.

BrianDang03: Added comments for the two for loops.

test/src/test_steqr.cpp
```diff
@@ -58,28 +58,34 @@ TEMPLATE_TEST_CASE("steqr is backward stable",
...
         // order
         for (idx_t i = 0; i < n - 1; ++i) {
-            CHECK(d[i] <= d[i + 1]);
+            CHECK((d1[i] == d2[i] && d2[i] <= d2[i + 1]));
```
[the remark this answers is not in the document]
BrianDang03: Created two loops of CHECKS

PR #485, opened 2025-07-07
Try trmm out

EllaAT01:
yeeyee

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -21,6 +21,15 @@ add_custom_command(
     "${CMAKE_CURRENT_BINARY_DIR}/geqr2/example_geqr2${CMAKE_EXECUTABLE_SUFFIX}"
     && echo ""
 )
+# add the example geqrtr
+add_subdirectory( geqrtr )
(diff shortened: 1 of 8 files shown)


test/src/test_bidiag_modified.cpp
```diff
@@ -0,0 +1,131 @@
+/// @file test_bidiag.cpp
```
langou, line 1: I do not think we want this "test_bidiag_modified.cpp" in the pull request.

test/src/CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ add_executable(test_householder_QR test_householder_QR.cpp)
 # add_executable(test_gerq2 test_gerq2.cpp)
 add_executable(test_trtri test_trtri.cpp)
 add_executable(test_bidiag test_bidiag.cpp)
+add_executable(test_bidiag_modified test_bidiag_modified.cpp)
```
langou, line 71: I do not think we want this "test_bidiag_modified.cpp" in the pull request.

PR #492, opened 2025-07-08
trmm_out request

EllaAT01:
only works for transA = noTrans + small fix in geqrtr ex.


test/src/test_trmm_out.cpp
```diff
@@ -0,0 +1,189 @@
...
+                C(i, j) -= C_copy(i, j);
+            }
+        }
+
+        real_t normC = lange(Norm::Fro, C);
+
+        normC = normC /
+                ((real(alpha) * normA * normB) + (real(beta) * normCbefore));
+
+        CHECK(normC <= tol);
```
langou, line 187: Change to:

real_t normC = lange(Norm::Fro, C);
CHECK(normC <= tol * ((real(alpha) * normA * normB) + (real(beta) * normCbefore)) );

the reason is that maybe alpha, normaA, normB, beta and normC are all zeros and it is not a good idea to divide zero by zero with

normC / ((real(alpha) * normA * normB) + (real(beta) * normCbefore))

It is better to multiply on the other side by zero

langou, line 187: With respect to CHECK(normC <= tol), the issue is that if the check fails and the precision is 1e-16 then we see something like 0.0 <= 0.0 because the precision of print of Catch2 is not much.

As of 2.8.0 there is a new way to do this but @luis-carlos-gutierrez reported that 2.8.0 does not work in the CI. For reference:
https://github.com/catchorg/Catch2/blob/devel/docs/tostring.md#floating-point-precision
and then something like:
        Catch::StringMaker<float>::precision = 15;
works great. (But not in the CI.)

One way to go around this is simply to write:
if ( !( normC <= tol ) ){ print normC and to to some accuracy }

PR #495, opened 2025-07-09
Try pbtrf fullaccess

kcunningturkey:
cholesky of banded

### DIFF
--- a/test/src/test_pbtrf_with_workspace.cpp
+++ b/test/src/test_pbtrf_with_workspace.cpp
@@ -53,14 +53,14 @@ TEMPLATE_TEST_CASE("triagular matrix-matrix multiplication is backward stable",
     // MatrixMarket reader
     MatrixMarket mm;
 
-    const idx_t n = GENERATE(13, 20, 25, 45, 131);
-    const idx_t kd = GENERATE(5, 7, 9, 12, 20);
+    const idx_t n = GENERATE(2, 13, 20, 25, 45, 131);
+    const idx_t kd = GENERATE(1, 5, 7, 9, 12, 20);
 
     srand(3);
 
     const Uplo uplo = GENERATE(Uplo::Upper, Uplo::Lower);
 
-    const size_t nb = GENERATE(1, 3, 4, 7, 9, 11);
+    const size_t nb = GENERATE(1, 2, 3, 4, 7, 9, 11);
 
     DYNAMIC_SECTION("n = " << n << " kd = " << kd << " Uplo = " << uplo
                            << " nb = " << nb)


include/tlapack/lapack/pbtrf_with_workspace.hpp
```diff
@@ -0,0 +1,223 @@
...
+     idx_t n = nrows(A);
+
+     std::vector<T> work_(nb * nb);
+     for (idx_t ii = 0; ii < nb * nb; ++ii) {
+         if constexpr (tlapack::is_complex<T>) {
+             work_[ii] = T(0, 0);
+         }
+         else
+             work_[ii] = 0;
+     }
```
langou, line 72: move these lines one line below and replace with 
laset( GENERAL, real_t(0), real_t(0), work);
I think this will be cleaner.

include/tlapack/lapack/pbtrf_with_workspace.hpp
```diff
@@ -0,0 +1,223 @@
...
+     using T = tlapack::type_t<matrix_t>;
+     using idx_t = tlapack::size_type<matrix_t>;
+     using range = tlapack::pair<idx_t, idx_t>;
+     using real_t = tlapack::real_type<T>;
+
+     tlapack::Create<matrix_t> new_matrix;
+
+     const idx_t nb = opts.nb;
+
+     idx_t n = nrows(A);
```
langou, line 63: add a few tlapack_check such as:
    tlapack_check(uplo == Uplo::Lower || uplo == Uplo::Upper);
    tlapack_check(nrows(A) == ncols(A));
    tlapack_check(kd <= nrows(A));

include/tlapack/lapack/pbtrf_with_workspace.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+     std::vector<T> work_(nb * nb);
```
langou, line 65: (COMMENT, NO ACTION REQUIRED, READ and CLICK "resolve")
Good with me to have memory allocation within the code for now.
Mostly because we have an algorithm to remove memory allocation. So great, this will go away.
If we keep a workspace code, then we will have to have a memory allocation layer. (See other \<T\>LAPACK codes.)

include/tlapack/lapack/pbtrf_with_workspace.hpp
```diff
@@ -0,0 +1,223 @@
...
+     auto work = new_matrix(work_, nb, nb);
+
+     if (uplo == tlapack::Uplo::Upper) {
+         for (idx_t i = 0; i < n; i += nb) {
+             idx_t ib = (n < nb + i) ? n - i : nb;
+
+             auto A00 = slice(A, range(i, min(ib + i, n)),
+                                  range(i, std::min(i + ib, n)));
+
+             potf2(tlapack::Uplo::Upper, A00);
```
langou, line 82: I do not know what we call here. It feels like we should call POTRF and let POTRF choose whether it wants POTF2 or something else. I think calling POTRF is better here.

test/src/test_pbtrf_with_workspace.cpp
```diff
@@ -0,0 +1,154 @@
...
+    srand(3);
+
+    const Uplo uplo = GENERATE(Uplo::Upper, Uplo::Lower);
+
+    const size_t nb = GENERATE(1, 3, 4, 7, 9, 11);
+
+    DYNAMIC_SECTION("n = " << n << " kd = " << kd << " Uplo = " << uplo
+                           << " nb = " << nb)
+    {
+        if (nb <= kd && kd < n) {
```
langou, line 67: I would like to let kd = n as a possibility. Please do: 
```cpp
if (nb <= kd && kd <= n) {
```
And in the GENERATE above have some n = kd. There is n = 20 and kd = 20, so that would work.

langou (commented):
Very nice! A few suggestions.

PR #559, opened 2026-06-05
Test geqr2

james-barton0819:
Creating a test for geqr2 (test_geqr2.cpp) and updating the MakeFile.


james-barton0819:
Good job!

PR #578, opened 2026-06-09
geqrt3 kernel + example

hchagerman:


include/tlapack/lapack/geqrt3.hpp
```diff
@@ -0,0 +1,156 @@
...
+        // no additional flops, just copy
+        tlapack::lacpy(tlapack::Uplo::General, A12, T12);
+
+        // step 3: A11T * T12 = T12
+
+        tlapack::trmm(tlapack::Side::Left, tlapack::Uplo::Lower,
+                      tlapack::Op::ConjTrans, tlapack::Diag::Unit,
+                      static_cast<T>(1.0), A11, T12);
+
+        // step 4: T12 + (A21T * A22) = T12
```
langou: Maybe use ᴴ instead of A21T. (1) use superscript, and (2) ᴴ is better than ᵀ since code can be in ℂ

langou: // step 4: T12 + (A21ᴴ * A22) = T12

PR #589, opened 2026-06-10
geqrt3

hchagerman:


include/tlapack/lapack/geqrt3.hpp
```diff
@@ -0,0 +1,148 @@
...
+        auto A22 = slice(A, range(m1, m2), range(n1, n));
+        auto A22_32 = slice(A, range(m1, m3), range(n1, n));
+        auto A31 = slice(A, range(m2, m3), range(0, n1));
+        auto A32 = slice(A, range(m2, m3), range(n1, n));
+        auto T11 = slice(Tmatrix, range(0, n1), range(0, n1));
+        auto T12 = slice(Tmatrix, range(0, n1), range(n1, n));
+        auto T22 = slice(Tmatrix, range(n1, n), range(n1, n));
+
+        // Cut down to one leading column
+        geqrt3<T>(A1, T11);
```
langou: Do you have to say `<T>` here? What is the difference between `geqrt3<T>(A1,T11)` and `geqrt3(A11,T11)`? I would prefer the latter if it works. Same comment for the second call to geqrt below.

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,249 @@
...
+        norm_orth_1 = tlapack::lange(tlapack::FROB_NORM, work);
+
+        if (verbose) {
+            std::cout << std::endl << "Q'Q-I = ";
+            printMatrix(work);
+        }
+    }
+
+    // 3) Compute ||QR - A||_F / ||A||_F
+    // A=q
```
langou: The `A=q` is weird here

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,249 @@
...
+        // Compute ||Q'Q - I||_F
+        norm_orth_1 = tlapack::lange(tlapack::FROB_NORM, work);
+
+        if (verbose) {
+            std::cout << std::endl << "Q'Q-I = ";
+            printMatrix(work);
+        }
+    }
+
+    // 3) Compute ||QR - A||_F / ||A||_F
```
langou: We could use our subscripts here: || QR - A ||ꜰ / || A ||ꜰ

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+        norm_repres_1 = tlapack::lange(tlapack::FROB_NORM, work) / normA;
+    }
+
+    // *) Output
+
+    std::cout << std::endl;
+    double seconds = elapsedQR.count() * 1.0e-9;
+    // Add the flops from geqrt3
+    double flopsQR = (3.0 * m * n * n) - ((5.0 / 6.0) * n * n * n) +
```
langou: Ah good. So ... this is interesting. You are taking the FLOPS count of GEQRT3 (which makes sense since you are using GEQRT3!) however another FLOPS count that could be interesting is the FLOPS count of GEQR2 ( 2 * m * n² - 2 / 3 * n³ ). It could be good to show both and find a name for both. I do not have an idea. 

Clearly the max perf you can get with the GEQR2 "FLOP rate" is 66% ( ≈ ( 3 * m * n² - 5 / 6 * n³ ) / ( 2 * m * n² - 2 / 3 * n³ )) but that's the reality of using GEQRT3. This is not really a FLOP rate but more of a "inverse of time" representation. 

I think showing both numbers would be good.

langou: The `3.0` casts the computation of `(3.0 * m * n * n)` to be done in floating-point arithmetic? Would it be `(3 * m * n * n)` that would be integer arithmetic and prone to integer overflow. I am worried that `3.0` is a little too subtle. I think it's fine. I would have typed: `(3.0 * ((double) m) * ((double) n) * ((double) m))` which is less subtle. (And same for all cubic terms.)

I do not know how \<T\>LAPACK is consistent with this. So maybe we leave it at that.

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,249 @@
...
+{
+    int m, n;
+
+    // Default arguments
+    m = (argc < 2) ? 1 : atoi(argv[1]);
+    n = (argc < 3) ? 1 : atoi(argv[2]);
+
+    srand(3);  // Init random seed
+
+    m = 64;
```
langou: I would really discourage showing a recursive code with `m` and `n` being power of 2's. It feels like the code only works for powers of 2's. I understand these are the default values and the user can use other values, but let us spice this recursion up. What about m = 73 and n= 55.

include/tlapack/lapack/geqrt3.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+        // step 2: Copy A12 into T12
+        // no additional flops, just copy
+        lacpy(Uplo::General, A12, T12);
+
+        // step 3: A11ᴴ * T12 = T12
```
langou: Can you please flip from 
`        // step 3: A11ᴴ * T12 = T12`
to
`        // step 3: T12 = A11ᴴ * T12`
?

I think this is easier to read. Same for all steps below.

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+    // Functors for creating new matrices
+    tlapack::Create<matrix_t> new_matrix;
+
+    // Turn it off if m or n are large
+    bool verbose = false;
+
+    // Arrays
+    std::vector<T> tau_work(n);
+    // create flops counter
```
langou: either do or don't.

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,249 @@
...
+        tlapack::laset(tlapack::GENERAL, static_cast<T>(0.0),
+                       static_cast<T>(1.0), work);
+        // work receives Q'Q - I
+        tlapack::gemm(tlapack::Op::ConjTrans, tlapack::Op::NoTrans,
+                      static_cast<T>(1.0), Q, Q, static_cast<T>(-1.0), work);
+
+        // Compute ||Q'Q - I||_F
+        norm_orth_1 = tlapack::lange(tlapack::FROB_NORM, work);
+
+        if (verbose) {
```
langou, line 166: I am confused with this `verbose` variable. This is a boolean. And the behavior is "weird". It feels it only triggers this print. I would suggest to remove `verbose`. If you want a `verbose` variable, I would leave all these print by default and maybe `verbose` ON would print the matrices or something like this. I think no `verbose` and no print of matrices is a good idea.

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,249 @@
...
+        ;
+
+        // work receives the identity n*n
+        tlapack::laset(tlapack::GENERAL, static_cast<T>(0.0),
+                       static_cast<T>(1.0), work);
+        // work receives Q'Q - I
+        tlapack::gemm(tlapack::Op::ConjTrans, tlapack::Op::NoTrans,
+                      static_cast<T>(1.0), Q, Q, static_cast<T>(-1.0), work);
+
+        // Compute ||Q'Q - I||_F
```
langou: We could use our subscripts and superscripts here: || Qᴴ Q - I ||ꜰ

examples/geqrt3/example_geqrt3.cpp
langou: Please discard these two comments. I see that you are printing the whole Qᴴ Q - I matrix. Then this is a good use of `verbose`. All good.

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,249 @@
...
+        norm_orth_1 = tlapack::lange(tlapack::FROB_NORM, work);
+
+        if (verbose) {
+            std::cout << std::endl << "Q'Q-I = ";
```
langou: We could use our superscripts here: Qᴴ Q - I

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+        // work receives the identity n*n
+        tlapack::laset(tlapack::GENERAL, static_cast<T>(0.0),
+                       static_cast<T>(1.0), work);
+        // work receives Q'Q - I
+        tlapack::gemm(tlapack::Op::ConjTrans, tlapack::Op::NoTrans,
+                      static_cast<T>(1.0), Q, Q, static_cast<T>(-1.0), work);
+
+        // Compute ||Q'Q - I||_F
+        norm_orth_1 = tlapack::lange(tlapack::FROB_NORM, work);
```
langou: I think the `_1` here means that you plan to use 1-norm. It is then weird to use the Frobenius norm. 
I do not know whether this is a copy-paste from another \<T\>LAPACK code but this is weird.
Maybe just call this variable `norm_orth` or `norm_orth_F`
Same comment for `norm_repres_1`.

PR #591, opened 2026-06-11
geqtr3

hchagerman:


examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,267 @@
...
+    // 2) Compute ||Qᴴ Q - I||ꜰ
+
+    {
+        // Copy Upper Triangle of A into R
+        tlapack::lacpy(tlapack::Uplo::Upper, Q, R);
+
+        // Copy the Householder vectors into V
+        tlapack::lacpy(tlapack::GENERAL, Q, V);
+
+        // Make Q the indentity matrix
```
langou, line 135: typo: identity

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+        norm_repres = tlapack::lange(tlapack::FROB_NORM, work) / normA;
+    }
+
+    // *) Output
+
+    std::cout << std::endl;
+    double seconds = elapsedQR.count() * 1.0e-9;
+
+    //(3*m*n² - 5/6*n³) + (1/2*n²) + (1/3*n))
```
langou, line 197: I think you want to remove the second and third order terms. So only keep cubic terms: (3*m*n² - 5/6*n³), remove the rest. Else, you would have to be consistent for the GEQR2_FLOPS below. The way GEQR2_FLOPS is computed is good. Only first order (cubic) terms. Same for GEQRT3.

examples/geqrt3/example_geqrt3.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+        // Copy the Householder vectors into V
+        tlapack::lacpy(tlapack::GENERAL, Q, V);
+
+        // Make Q the indentity matrix
+        for (idx_t j = 0; j < n; ++j)
+            for (idx_t i = 0; i < m; ++i)
+                Q(i, j) = static_cast<T>(0.0);
+
+        for (idx_t j = 0; j < std::min(m, n); ++j)
```
langou, line 140: I assume laset does not work for a rectangular matrix? (I do not know.) If we can use laset to set the Identity matrix in Q here, this might be cleaner.

include/tlapack/lapack/geqrt3.hpp
```diff
@@ -0,0 +1,142 @@
...
+        auto A21 = slice(A, range(m1, m2), range(0, n1));
+        auto A22 = slice(A, range(m1, m2), range(n1, n));
+        auto A22_32 = slice(A, range(m1, m3), range(n1, n));
+        auto A31 = slice(A, range(m2, m3), range(0, n1));
+        auto A32 = slice(A, range(m2, m3), range(n1, n));
+        auto T11 = slice(Tmatrix, range(0, n1), range(0, n1));
+        auto T12 = slice(Tmatrix, range(0, n1), range(n1, n));
+        auto T22 = slice(Tmatrix, range(n1, n), range(n1, n));
+
+        // Cut down to one leading column
```
langou, line 73: Maybe something like "recurse on the left" would be better than "Cut down to one leading column"

langou: Or         
// step 1: Compute the QR factorization of A1

include/tlapack/lapack/geqrt3.hpp
```diff
@@ -0,0 +1,142 @@
...
+        trmm(Side::Left, Uplo::Lower, Op::NoTrans, Diag::Unit, T(1.0), A11,
+             T12);
+
+        // step 8: A12 = A12 - T12
+        for (idx_t j = 0; j < n2; ++j) {
+            for (idx_t i = 0; i < m1; ++i) {
+                A12(i, j) -= T12(i, j);
+            }
+        }
+        // step 9: Compute the QR factorization of T22
```
langou, line 112: Not T22, you mean A22_32

PR #603, opened 2026-06-16
Rgeqrf pull request

james-barton0819:


examples/rgeqrf/example_rgeqrf.cpp
```diff
@@ -0,0 +1,267 @@
...
+        tlapack::rgeqrf(Q, Tmatrix);
+    }
+    // Record end time
+    auto endQR = std::chrono::high_resolution_clock::now();
+
+    // Compute elapsed time in nanoseconds
+    auto elapsedQR =
+        std::chrono::duration_cast<std::chrono::nanoseconds>(endQR - startQR);
+
+    // Compute FLOPS **WIP** Needs to be calculated
```
langou: Correct, the FLOPS needs to be calculated! 
First order only is enough (so just cubic terms).

langou (approved):

PR #604, opened 2026-06-17
Improve qz eig22

thijssteel:
lahqz_eig22 calculates the eigenvalues of a 2x2 generalized eigenvalue block.

In the previous version, this was done by inverting the B matrix and then using the standard eigenvalue version, which obviously caused issues for singular matrices.

This version translates dlag2 from LAPACK to C++, with two modifications:
1. Instead of perturbing singular matrices to avoid issues, we handle these matrices separately by deflating the infinite eigenvalue.
2. This version can also handle 2x2 blocks of complex valued blocks.

I also added lahqz_schur22, which calculates the schur factorisation. The most important thing here is that we make sure that any 2x2 block with real eigenvalues is reduced to two 1x1 blocks. I have some suspicions that this is important for the schur swapping.


test/src/test_qz_eig22.cpp
```diff
@@ -31,7 +31,7 @@ TEMPLATE_TEST_CASE(
     using T = type_t<matrix_t>;
     using idx_t = size_type<matrix_t>;
     using real_t = real_type<T>;
-    using complex_t = complex_type<real_t>;
+    using complex_t = complex_type<T>;
```
weslleyspereira, line 34: Oops... This is kind of issue that I might be able to help (and it is probably my fault). Let me know how that goes...

thijssteel: Any help would be appreciated, I currently have no clue why It won't compile on the CI. Everything works fine locally...

thijssteel: I'm just going to disable that branch it for now. That seems to fix it.

PR #608, opened 2026-06-23
Improve generalized schur swap

thijssteel:
Improved generalized schur swap in the following ways:

Improved stability of swapping 1x1 with 2x2 blocks by trying both options.

Added iterative refinement for swapping 2x2 with 2x2 blocks, allowing ill-conditioned blocks with close eigenvalues to still be swapped accurately (in most cases, it may take too many iterations for some blocks)

Always use the full normalization using schur22 instead of only apply the svd normalisation. Slightly rewrote the interface of schur22 to facilitate that.

I occasionally got large errors which could be attributed to schur22. I updated one of the selection criteria there which significantly reduced the error. I will definitely look into pushing that change to LAPACK.

Ready for review now if someone want to check.

### DIFF
--- a/include/tlapack/lapack/generalized_schur_swap.hpp
+++ b/include/tlapack/lapack/generalized_schur_swap.hpp
@@ -347,18 +347,23 @@ int generalized_schur_swap(bool want_q,
         //
         std::vector<T> H_;
         auto H = new_matrix(H_, 2, 3);
-        std::vector<T> v(3);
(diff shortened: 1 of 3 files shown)


thijssteel:
[report.pdf](https://github.com/user-attachments/files/29297944/report.pdf)
I wrote a short report to compare the two methods. Failures appear to be much much rarer now!

PR #609, opened 2026-06-23
nested for-loops to axpy call update

james-barton0819:


langou:
One comment with respect to this PR. For a software like <T>LAPACK, where we do not assume anything on the matrix data layout, it would be make sense to have a function to perform A = A - B (for example), that'd be GEADD. If the matrix is stored row-major, you would want to stream along the rows. If the matrix is stored column-major, you would want to stream along the columns. Etc. In a high level routine like RGEQRF, you do not want to make assumption on how the data is stored, and you do not want to start messing up with this, so you call GEADD and hopefully GEADD would have optimized versions depending on the data layout.

PR #615, opened 2026-06-25
added larft recursive

hchagerman:


langou:
To test this larft_recursive, it should be possible to replace all the calls to larft in 
- gelqf.hpp
- gelqt.hpp
- geqlf.hpp
- geqrf.hpp
- gerqf.hpp
- ungq.hpp 
- unmq.hpp 
to calls to larft_recursive and then check that the test suite completes correctly.

I am not so sure how to do this with our test suite in a nice way. I do not really want the test suite to check all the variants of larft for the whole \<T\>LAPACK, but it would be a good check locally on a laptop.

PR #616, opened 2026-06-25
geqrt3 nx attribute

james-barton0819:
Created nx attribute for geqrt3.hpp and updated the example and test files for rgeqrf and geqrt3 to utilize it

### DIFF
--- a/include/tlapack/lapack/geqrt3.hpp
+++ b/include/tlapack/lapack/geqrt3.hpp
@@ -15,17 +15,23 @@
 
 #include "tlapack/blas/gemm.hpp"
 #include "tlapack/blas/trmm.hpp"
+#include "tlapack/lapack/geqrf.hpp"
 #include "tlapack/lapack/lacpy.hpp"
 #include "tlapack/lapack/larfg.hpp"
+#include "tlapack/lapack/larft.hpp"
 
 namespace tlapack {
 /**
  * By toggling isw to true, the geqrt3 routine will stop its loop before
  * computing the the upper right block of the T matrix, saving on the potential
  * FLOP count.
+ *
+ * Changing the value of nx will limit how few columns the recursion goes to
+ * before returning.
  */
 struct Geqrt3Opts {
     bool isw = false;
+    size_t nx = 1;
 };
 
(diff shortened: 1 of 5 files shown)


include/tlapack/lapack/geqrt3.hpp
```diff
@@ -72,6 +78,13 @@ void geqrt3(matrix_a& A, matrix_h& Tmatrix, const Geqrt3Opts& opts = {})
         // Populate matrix T with an elementary reflector
         larfg(Direction::Forward, StoreV::Columnwise, a_vector, Tmatrix(0, 0));
     }
+    else if (n <= opts.nx) {
+        auto tau = diag(Tmatrix);
```
langou, line 82: That's a nice use of the "diag" view. Good to see that this is useful. Very nice.

include/tlapack/lapack/geqrt3.hpp
```diff
@@ -72,6 +78,13 @@ void geqrt3(matrix_a& A, matrix_h& Tmatrix, const Geqrt3Opts& opts = {})
...
+
+        geqrf(A, tau);
+
+        larft(Direction::Forward, StoreV::Columnwise, A, tau, Tmatrix);
```
langou, line 86: It is interesting to see this. So tau and the diagonal of Tmatrix are aliased. I do not think that I have ever seen this. Interesting. I think this is fine.

langou (approved):

include/tlapack/lapack/geqrt3.hpp
langou: I think a better interface for LARFT would be to ask the user to set tau on the diagonal of T. That would prevent the aliasing (or the duplication of two tau's.) No need to do this. This is a general comment.

Note that the aliasing of TAU with the diagonal of T is not possible in LAPACK because tau does not have an increment. It would be possible to pass TAU as the diagonal of T in LAPACK if we had an increment for TAU. And then we would set TAU to point to T and INC_TAU = LDT+1 and done. But as LARFT, we cannot do this in LAPACK.

